PR #35, opened 2022-03-18
Add tests to os_info

harlem88:
- Check the behaviour if os_release contains an empty line in the middle.
- Check the behaviour if os_release contains only the os_name.
- Check the behaviour if os_release file is malformed.
- Split parse_os_info function.
- Separate key_value parsing from osName and osVersion.
- Add tests to parse_key_value line function.

### DIFF
--- a/src/telemetry/os_info.rs
+++ b/src/telemetry/os_info.rs
@@ -100,4 +100,32 @@ BUG_REPORT_URL="https://bugs.debian.org/""#;
         assert_eq!(data["/osName"], "Debian GNU/Linux");
         assert_eq!(data["/osVersion"], "11");
     }
+
+    #[test]
+    fn os_release_parsing_with_middle_empty_line() {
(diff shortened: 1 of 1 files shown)


src/telemetry/os_info.rs
```diff
@@ -100,4 +100,46 @@ BUG_REPORT_URL="https://bugs.debian.org/""#;
...
+    }
+
+    #[test]
+    fn parse_key_value_line_good() {
+        let line = r#"OS="Arch"#;
+
+        let (key, value) = parse_key_value_line(line).unwrap();
+        assert_eq!(key, "OS");
+        assert_eq!(value, "Arch");
+    }
```
FrancescoVaiani, line 144: Previous tests are named `os_release_with_only_name` and `os_release_malformed`. Could these two be called `parse_key_value_line_malformed` and `parse_key_value_line` for consistency?

FrancescoVaiani (approved):

PR #44, opened 2022-03-30
Improve OTA handler

harlem88:
- Move OTA stuffs in ota module folder.
- Add `AstarteClient` struct and tests to `OTAHandler`.
Wrap `AstarteSDK` is useful both for not depending directly on the SDK and for testing reasons.
- Add `mockall` library.


src/astarte.rs
```diff
@@ -0,0 +1,67 @@
...
+pub trait Publisher: Send + Sync {
+    async fn send_object<T: 'static>(
+        &self,
+        interface_name: &str,
+        interface_path: &str,
+        data: T,
+    ) -> Result<(), AstarteError>
+    where
+        T: serde::Serialize + Send;
+}
```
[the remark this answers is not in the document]
rbino: Also, the trait should probably support additional methods (e.g. `send_object_with_timestamp` and `send`). Are they going to be added in future PRs? In that case it would probably be nice to leave a TODO

src/ota/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+#[async_trait]
+impl<'a> OTA for OTARauc<'a> {
+    async fn install_bundle(&self, _source: &str) -> Result<(), DeviceManagerError> {
+        self.rauc
+            .install_bundle("path", std::collections::HashMap::new())
+            .await?;
+        Ok(())
+    }
```
rbino, line 55: Let's try to fix the Codecov configuration because reviewing PRs interspersed with all these warnings is a little painful

rbino (changes requested):

src/ota/ota_handler.rs
```diff
@@ -68,7 +76,7 @@ impl OTAStatus {
 }
 
 pub struct OTAHandler<'a> {
-    rauc: RaucProxy<'a>,
+    ota: Box<dyn OTA + 'a>,
```
Annopaolo, line 79: Could  `impl OTA` be used here? At the moment, it seems only `OTARauc` is being used (cf. https://users.rust-lang.org/t/difference-between-returning-dyn-box-trait-and-impl-trait/57640)

src/ota/ota_handler.rs
```diff
@@ -314,4 +307,171 @@ mod tests {
...
+            state_repository: Box::new(state_mock),
+            download_file_path: "".to_owned(),
+        };
+
+        let result = block_on(ota_handler.handle_ota_event(&publisher, "", ""));
+        assert!(result.is_err());
+
+        let result_error = get_string_value_from_update_error(result.err().unwrap());
+        let error_expected = "bundle is not compatible";
+        assert_eq!(error_expected, result_error);
```
Annopaolo, line 355: Could this be made clearer by explicitly letting `get_string_value_from_update_error` fail if `result.err().unwrap()` is not an `UpdateError`?

Annopaolo (changes requested):

rbino (approved):

Annopaolo (approved):

PR #47, opened 2022-04-05
Implement io.edgehog.devicemanager.StorageUsage

nanjarapalli:


src/telemetry/storage_usage.rs
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use crate::error::DeviceManagerError;
+use astarte_sdk::types::AstarteType;
+use std::collections::HashMap;
+use sysinfo::{DiskExt, System, SystemExt};
+
+/// get structured data for `io.edgehog.devicemanager.StorageUsage` interface
+pub fn get_storage_usage() -> Result<Vec<HashMap<String, AstarteType>>, DeviceManagerError> {
```
shikoku1, line 27: You can use a struct instead of an HashMap, since StorageUsage is an object, similar to get_system_status

harlem88:
@nanjarapalli you should add also the first send of storage_usage here https://github.com/edgehog-device-manager/edgehog-device-runtime/blob/5ea2fa76246833ae6ce707a06d04a9dc43dd397f/src/lib.rs#L121.

src/telemetry/storage_usage.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/// get available partition disks on the system
+fn get_disks() -> Result<Vec<DiskUsage>, DeviceManagerError> {
+    let mut ret: Vec<DiskUsage> = Vec::new();
+    let mut sys = System::new_all();
+    sys.refresh_disks();
+
+    for disk in sys.disks() {
+        let mut map: HashMap<String, AstarteType> = HashMap::new();
```
shikoku1, line 42: is this variable used?

src/telemetry/storage_usage.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/// get structured data for `io.edgehog.devicemanager.StorageUsage` interface
+pub fn get_storage_usage() -> Result<HashMap<String, AstarteType>, DeviceManagerError> {
+    let mut ret: HashMap<String, AstarteType> = HashMap::new();
+
+    let disks = get_disks().unwrap();
+
+    for disk in disks.iter() {
+        println!(" {:?}", disk.disk_name);
```
shikoku1, line 68: leftover? 
please use macros in the `log` crate if you want to do logging

src/telemetry/storage_usage.rs
```diff
@@ -0,0 +1,79 @@
...
+        });
+    }
+    Ok(ret)
+}
+
+/// get structured data for `io.edgehog.devicemanager.StorageUsage` interface
+pub fn get_storage_usage() -> Result<HashMap<String, AstarteType>, DeviceManagerError> {
+    let mut ret: HashMap<String, AstarteType> = HashMap::new();
+
+    let disks = get_disks().unwrap();
```
shikoku1, line 65: do not `.unwrap()` unless necessary, please handle or propagate this error

src/telemetry/storage_usage.rs
```diff
@@ -0,0 +1,79 @@
...
+            disk_name: disk_name.to_owned() as String,
+            total_bytes: disk.total_space() as i64,
+            free_bytes: disk.available_space() as i64,
+        });
+    }
+    Ok(ret)
+}
+
+/// get structured data for `io.edgehog.devicemanager.StorageUsage` interface
+pub fn get_storage_usage() -> Result<HashMap<String, AstarteType>, DeviceManagerError> {
```
shikoku1, line 62: You can use a struct instead of an HashMap, since StorageUsage is an object, similar to get_system_status

src/telemetry/storage_usage.rs
```diff
@@ -0,0 +1,79 @@
...
+        let mut map: HashMap<String, AstarteType> = HashMap::new();
+        let disk_name = disk.name().to_str().unwrap().replace("/dev", "");
+        map.insert(
+            format!("{}/totalBytes", disk_name),
+            (disk.total_space() as i64).into(),
+        );
+        map.insert(
+            format!("{}/freeBytes", disk_name),
+            (disk.available_space() as i64).into(),
+        );
```
harlem88, line 51: Remove this part, you have already converted the Disk Usage to Astarte properties in `get_storage_usage` function, also the map definition at 42 line.

nanjarapalli: removed

src/telemetry/storage_usage.rs
nanjarapalli: Struct is used in getDisks already and here Hashmap is required to format as per astarte interface

src/telemetry/storage_usage.rs
nanjarapalli: propagate the error

src/telemetry/storage_usage.rs
nanjarapalli: remoevd

src/telemetry/storage_usage.rs
nanjarapalli: not used and removed

src/telemetry/storage_usage.rs
nanjarapalli: switched to struct and used in getDisks method

src/telemetry/storage_usage.rs
shikoku1: please look at get_system_status

nanjarapalli: I looked already into get_system_status. It is not similar. I cannot use a struct here because I need to construct dynamic field names at runtime inside the struct.

src/telemetry/storage_usage.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/// get structured data for `io.edgehog.devicemanager.StorageUsage` interface
+pub fn get_storage_usage() -> Result<HashMap<String, DiskUsage>, DeviceManagerError> {
+    let mut ret: HashMap<String, DiskUsage> = HashMap::new();
+    let mut sys = System::new_all();
+    sys.refresh_disks();
+
+    for disk in sys.disks() {
+        let disk_name = disk.name().to_str().unwrap().replace("/dev", "");
```
harlem88, line 40: Remove the last `/` because it is not part of the disk name, but remember to add `/` 
as **interface_path** argument of the **send_object** function before send the object to Astarte.
```rust
replace("/dev/", "");
```

harlem88 (changes requested):

harlem88 (approved):

FrancescoVaiani (approved):

src/telemetry/storage_usage.rs
```diff
@@ -0,0 +1,50 @@
...
+use sysinfo::{DiskExt, System, SystemExt};
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DiskUsage {
+    pub total_bytes: i64,
+    pub free_bytes: i64,
+}
+
+/// get structured data for `io.edgehog.devicemanager.StorageUsage` interface
```
Annopaolo, line 33: Maybe be some more descriptive? E.g. why is `"/dev/"` being removed from disk name?

Annopaolo (changes requested):

src/telemetry/storage_usage.rs
nanjarapalli: added

bettio (approved):

PR #48, opened 2022-04-06
Retrieve hardware id from dbus

FrancescoVaiani:
- Add an example project to demonstrate how the `hardware id` should be published on the dbus. In this example, the id is taken from the BIOS standard locations as for DMI specifications. This only works on x86/x64 machines.
- Retrieve the `hardware id` form the dbus.
- Use the `hardware id` retrived from the dbus to generate a valid Astarte `device id`
- Make the `device_id` and `credentials_secret` CI params optionals.
- Add the `pairing_token` CI optional parameter to allow device registration


dmi_id_dbus_service/src/main.rs
```diff
@@ -0,0 +1,76 @@
...
+            .to_string();
+        if data != "" {
+            return Some(data);
+        }
+    }
+    None
+}
+
+// Simple DBUS service that retrieves a machine specific id and publish it on a system channel
+#[async_std::main]
```
shikoku1, line 60: Is there a reason to use async_std? We are using tokio for everything else

FrancescoVaiani: You are right, I forgot about it. Fixed

dmi_id_dbus_service/src/main.rs
```diff
@@ -0,0 +1,76 @@
...
+    #[dbus_interface(property)]
+    async fn device_id(&self) -> &str {
+        &self.id
+    }
+}
+
+// Get hardware id from board serial or from product uuid provided by the BIOS (standard SMBIOS/DMI).
+fn get_hardware_id() -> Option<String> {
+    let mut data = fs::read_to_string("/sys/class/dmi/id/board_serial")
+        .expect("Unable to read file")
```
shikoku1, line 42: We should try to propagate errors instead of panicking

dmi_id_dbus_service/src/main.rs
```diff
@@ -0,0 +1,76 @@
...
+        }
+    }
+    None
+}
+
+// Simple DBUS service that retrieves a machine specific id and publish it on a system channel
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let hardware_id: String = match get_hardware_id() {
+        None => "".to_string(),
```
shikoku1, line 63: is this a manual version of unwrap_or_default()?

dmi_id_dbus_service/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+        &self.id
+    }
+}
+
+// Get hardware id from board serial or from product uuid provided by the BIOS (standard SMBIOS/DMI).
+fn get_hardware_id() -> Option<String> {
+    return match fs::read_to_string("/sys/class/dmi/id/board_serial"){
+        Err(_e) => {
+            match fs::read_to_string("/sys/class/dmi/id/product_uuid"){
+                Err(_e) => None,
```
shikoku1, line 44: is this a manual version of .ok()?

dmi_id_dbus_service/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+                Ok(_data) => {
+                    if _data != "" {
```
shikoku1, line 46: no need for the `_` if the variable is used

src/lib.rs
```diff
@@ -145,3 +158,52 @@ impl DeviceManager {
         Ok(())
     }
 }
+
+async fn get_device_id(opt_device_id: Option<String>) -> Result<String, DeviceManagerError> {
+    return if opt_device_id != None {
+        Ok(opt_device_id.unwrap())
```
shikoku1, line 164: Is this a manual version of `if let`?
Also, you can delete the `return if` and remove the final ;

dmi_id_dbus_service/src/main.rs
```diff
@@ -0,0 +1,60 @@
...
+    if data.is_empty() {
+        data = fs::read_to_string("/sys/class/dmi/id/product_uuid")?;
+    }
+    Ok(data)
+}
+
+// Simple DBUS service that retrieves a machine specific id and publish it on a system channel
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let hardware_id: String = get_hardware_id().unwrap_or_default();
```
shikoku1, line 49: I don't think it's necessary to specify the type in this case

src/lib.rs
```diff
@@ -145,3 +158,52 @@ impl DeviceManager {
...
+        let registration = registration::register_device(
+            &opts.pairing_token.clone().unwrap(),
+            &opts.pairing_url,
+            &opts.realm,
+            &device_id,
+        )
+        .await;
+        match registration {
+            Err(_e) => return Err(DeviceManagerError::FatalError("Pairing error".to_string())),
+            Ok(_credential_secret) => {
```
shikoku1, line 198: why the _ if you are using the variables, it's not necessary

src/lib.rs
harlem88: It is also available the `is_none` function :grin: https://doc.rust-lang.org/std/option/enum.Option.html#method.is_none

harlem88: Sure, but I only write because `option != None` was horrible to see. :grinning:

src/lib.rs
```diff
@@ -145,3 +158,52 @@ impl DeviceManager {
...
+
+async fn get_device_id(opt_device_id: Option<String>) -> Result<String, DeviceManagerError> {
+    return if opt_device_id != None {
+        Ok(opt_device_id.unwrap())
+    } else {
+        let connection = zbus::Connection::system().await?;
+        let proxy = DeviceProxy::new(&connection).await?;
+        let namespace = Uuid::parse_str("f79ad91f-c638-4889-ae74-9d001a3b4cf8").unwrap();
+        let hardware_id: String = proxy.device_id().await?;
+        if hardware_id == "" {
```
harlem88, line 170: Use `is_empty()` function.

`if hardware_id.is_empty() {`

harlem88 (changes requested):

src/lib.rs
```diff
@@ -145,3 +158,52 @@ impl DeviceManager {
...
+        Ok(registration::generate_uuid(namespace, &hardware_id))
+    }
+}
+
+async fn get_credentials_secret(
+    device_id: &str,
+    opts: &DeviceManagerOptions,
+) -> Result<String, DeviceManagerError> {
+    if opts.credentials_secret.is_some() {
+        Ok(opts.credentials_secret.clone().unwrap())
```
shikoku1, line 184: same as before, this is an `if let` done manually, it's a good practice to avoid calling `.unwrap()` and variants unless absolutely necessary

bettio (changes requested):

dmi_id_dbus_service/src/main.rs
```diff
@@ -0,0 +1,60 @@
...
+impl Device {
+    #[dbus_interface(property)]
+    async fn device_id(&self) -> &str {
+        &self.id
+    }
+}
+
+// Get hardware id from board serial or from product uuid provided by the BIOS (standard SMBIOS/DMI).
+fn get_hardware_id() -> io::Result<String> {
+    let mut data = fs::read_to_string("/sys/class/dmi/id/board_serial").unwrap_or_default();
```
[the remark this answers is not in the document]
FrancescoVaiani: Just a previous convention. That is the place where SECO stores their board number, but we can swap the order without problems.

bettio:
Two naming points here:
- I'm not sure if I would use the term `device_id` or `hardware_id` here, but then I would use it consistently.
- Also I would call this service `device-id-service` or `device-id-helper` (or `hardware-id-service` or `hardware-id-helper`)
Maybe @rbino has some points too.

Also I think on the longer term we should support 1 or 2 additional (and quite standard) alternatives, like taking it from `/proc/cpuinfo` (which is quite common) and taking it from boot parameters (I think they can be retrieved from `/proc/cmdline`). We should also allow to use a configuration file to tell the service which ID provider should be used.

Also I think the service should always return an UUID string, and we should hash it using a known namespace or whatever (let's also discuss this).

bettio:
One additional comment: let's always use as first commit the `cargo new` output with no customization. That commit should contain as commit message `cargo new FOOBAR`.

dmi-id-dbus-service/src/main.rs
```diff
@@ -0,0 +1,60 @@
...
+
+use std::fs;
+use std::{error::Error, io};
+use zbus::{dbus_interface, ConnectionBuilder};
+
+struct Device {
+    id: String,
+}
+
+#[dbus_interface(name = "io.edgehog.Device1")]
```
harlem88, line 29: Is it right the `1` at the end of name?

src/lib.rs
```diff
@@ -169,3 +178,46 @@ impl DeviceManager {
...
+    if let Some(secret) = opts.credentials_secret.clone() {
+        Ok(secret)
+    } else if Path::new(&format!("./{}.json", device_id)).exists() {
+        let reader = File::open(&format!("./{}.json", device_id)).unwrap();
+        Ok(serde_json::from_reader(reader).expect("Unable to read secret"))
+    } else if let Some(token) = opts.pairing_token.clone() {
+        let registration =
+            registration::register_device(&token, &opts.pairing_url, &opts.realm, &device_id).await;
+        match registration {
+            Err(_e) => return Err(DeviceManagerError::FatalError("Pairing error".to_string())),
```
shikoku1, line 211: you can map the error and then `?` it, instead of doing a whole `match`

FrancescoVaiani: the PairingError is not publicly exposed by the SDK

dmi-id-dbus-service/src/main.rs
```diff
@@ -0,0 +1,117 @@
...
+// Simple DBUS service that retrieves a machine specific id and publish it on a system channel
+#[tokio::main]
+async fn main() -> zbus::Result<()> {
+    let Cli {
+        file_path,
+        seco_dmi_serial,
+        kernel_cmdline_key,
+    } = Parser::parse();
+
+    if file_path.is_none() & kernel_cmdline_key.is_none() & !seco_dmi_serial.unwrap_or(false) {
```
harlem88, line 88: Why here you use the Bitwise AND and not the `&&` logical AND?

src/lib.rs
```diff
@@ -171,3 +180,45 @@ impl DeviceManager {
...
+            let writer = File::create(&format!("./{}.json", device_id)).unwrap();
+            serde_json::to_writer(writer, &credential_secret).expect("Unable to write secret");
+            Ok(credential_secret)
+        } else {
+            Err(DeviceManagerError::FatalError("Pairing error".to_string()))
+        }
+    } else {
+        return Err(DeviceManagerError::FatalError(
+            "Not enough arguments".to_string(),
+        ));
```
harlem88, line 222: Remove `return` and transform it to an expression, like the others above. 
``` rust
Err(DeviceManagerError::FatalError(
    "Not enough arguments".to_string(),
))
```

harlem88 (changes requested):

dmi-id-dbus-service/io.edgehog.Device.conf
```diff
@@ -0,0 +1,33 @@
...
+
+<!DOCTYPE busconfig PUBLIC
+ "-//freedesktop//DTD D-BUS Bus Configuration 1.0//EN"
+ "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
+<busconfig>
+
+  <!-- Only root can own the service -->
+  <policy user="root">
+    <allow own="io.edgehog.Device"/>
+    <allow send_destination="org.secomind.Device"/>
```
bettio, line 17: `io.edgehog`

dmi-id-dbus-service/io.edgehog.Device.conf
```diff
@@ -0,0 +1,33 @@
...
+    <allow send_interface="org.secomind.Device"/>
```
bettio, line 18: `io.edgehog`

dmi-id-dbus-service/io.edgehog.Device.conf
```diff
@@ -0,0 +1,33 @@
...
+  </policy>
+
+  <policy at_console="true">
+    <deny own="io.edgehog.Device"/>
+    <allow send_destination="org.secomind.Device"/>
```
bettio, line 23: `io.edgehog`

dmi-id-dbus-service/io.edgehog.Device.conf
```diff
@@ -0,0 +1,33 @@
...
+    <allow send_interface="org.secomind.Device"/>
```
bettio, line 24: `io.edgehog`

dmi-id-dbus-service/io.edgehog.Device.conf
```diff
@@ -0,0 +1,33 @@
...
+  </policy>
+
+  <policy context="default">
+    <deny own="io.edgehog.Device"/>
+    <allow send_destination="org.secomind.Device"/>
```
bettio, line 29: `io.edgehog`

dmi-id-dbus-service/io.edgehog.Device.conf
```diff
@@ -0,0 +1,33 @@
...
+    <allow send_interface="org.secomind.Device"/>
```
bettio, line 30: `io.edgehog`

dmi-id-dbus-service/io.edgehog.Device.conf
```diff
@@ -0,0 +1,33 @@
...
+#  SPDX-License-Identifier: Apache-2.0
+
+<?xml version="1.0" encoding="UTF-8"?> <!-- -*- XML -*- -->
+
+<!DOCTYPE busconfig PUBLIC
+ "-//freedesktop//DTD D-BUS Bus Configuration 1.0//EN"
+ "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
+<busconfig>
+
+  <!-- Only root can own the service -->
```
bettio, line 14: Let's use everywhere: `io.edgehog`

dmi-id-dbus-service/src/main.rs
```diff
@@ -0,0 +1,117 @@
...
+
+#[derive(Debug, Parser)]
+struct Cli {
+    // Retrieve hardware id from file
+    #[clap(short, long, conflicts_with_all=&["seco-dmi-serial","kernel-cmdline-key"])]
+    file_path: Option<String>,
+
+    // Shortcut per file-path = "/sys/class/dmi/id/board_serial"
+    #[clap(short, long, takes_value = false, conflicts_with_all=&["file-path","kernel-cmdline-key"])]
+    seco_dmi_serial: bool,
```
bettio, line 38: This is not specific to SECO boards, so let's rename it to just `dmi_serial`. Also since it is a boolean a meaningful verb should be used, like `use`, therefore: `use_dmi_serial` might be an option.

dmi-id-dbus-service/src/main.rs
```diff
@@ -0,0 +1,117 @@
...
+use uuid::Uuid;
+use zbus::{dbus_interface, ConnectionBuilder};
+
+const DMI_SERIAL_FILE_PATH: &str = "/sys/class/dmi/id/board_serial";
+const DEFAULT_NAMESPACE: &str = "f79ad91f-c638-4889-ae74-9d001a3b4cf8";
+
+#[derive(Debug, Parser)]
+struct Cli {
+    // Retrieve hardware id from file
+    #[clap(short, long, conflicts_with_all=&["seco-dmi-serial","kernel-cmdline-key"])]
```
bettio, line 33: This is not specific to SECO boards, so let's rename it to just `dmi_serial`.

Furthermore since we are using a boolean type a meaningful verb should be used, such as `use`, therefore: `use-dmi-serial` might be an option.

Let's rename everything accordingly.

dmi-id-dbus-service/README.md
```diff
@@ -0,0 +1,41 @@
+<!---
+  Copyright 2022 SECO Mind Srl
+
+  SPDX-License-Identifier: Apache-2.0
+-->
+
+# Hardware ID dbus Service Example
+Example project of a service that retrieves a hardware id from the BIOS (standard SMBIOS/DMI)
```
bettio, line 8: Let's call this a "Reference" rather than an "Example".

dmi-id-dbus-service/README.md
```diff
@@ -0,0 +1,41 @@
...
+and publishes it on a dbus channel.
```
bettio, line 9: that either fetches hardware id from SMBIOS/DMI or from the kernel command line.

dmi-id-dbus-service/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Setup
+Copy `org.secomind.Device.conf` in the dbus system configuration directory
```
bettio, line 12: io.edgehog

bettio (changes requested):

harlem88 (approved):

hardware-id-service/README.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+Build with cargo
+```bash
+cargo build --release
+```
+
+Run as superuser
+```bash
+sudo target/release/hardware_id_poc
```
bettio, line 27: Are we still calling it something_poc?

src/lib.rs
```diff
@@ -171,3 +180,45 @@ impl DeviceManager {
...
+        if let Ok(credential_secret) = registration {
+            let writer = File::create(&format!("./{}.json", device_id)).unwrap();
+            serde_json::to_writer(writer, &credential_secret).expect("Unable to write secret");
+            Ok(credential_secret)
+        } else {
+            Err(DeviceManagerError::FatalError("Pairing error".to_string()))
+        }
+    } else {
+        Err(DeviceManagerError::FatalError(
+            "Not enough arguments".to_string(),
```
bettio, line 221: `Not enough arguments` -> `Missing arguments`.

bettio (changes requested):

bettio (approved):

PR #60, opened 2022-05-04
System Info support

### LINKED ISSUE #9: Implement `io.edgehog.devicemanager.SystemInfo` Interface in Device Runtime

FrancescoVaiani:
Add `io.edgehog.devicemanager.SystemInfo` to `send_initial_telemetry`
with data fetched from the kernel cmdline and from ENV variables.

Close #9

### DIFF
--- a/src/telemetry/system_info.rs
+++ b/src/telemetry/system_info.rs
@@ -1,0 +1,73 @@
+/*
+ * This file is part of Edgehog.
+ *
+ * Copyright 2022 SECO Mind Srl
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+use crate::error::DeviceManagerError;
+use astarte_sdk::types::AstarteType;
+use procfs::cmdline;
+use std::collections::HashMap;
+use std::env;
+
+pub fn get_system_info() -> Result<HashMap<String, AstarteType>, DeviceManagerError> {
+    let cmdline_params = cmdline().unwrap();
+    let mut seral_number: Option<String> = env::var("EDGEHOG_SYSTEM_SERIAL_NUMBER").ok();
+    let mut part_number: Option<String> = env::var("EDGEHOG_SYSTEM_PART_NUMBER").ok();
+    for param in cmdline_params.iter() {
+        if seral_number.is_none() && param.starts_with("edgehog_system_serial_number") {
+            let first_half = format!("{}=", "edgehog_system_serial_number");
+            seral_number = Some(param.replace(first_half.as_str(), ""));
+        }
+        if part_number.is_none() && param.starts_with("edgehog_system_part_number") {
+            let first_half = format!("{}=", "edgehog_system_part_number");
+            part_number = Some(param.replace(first_half.as_str(), ""));
+        }
+    }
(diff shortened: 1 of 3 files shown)


system-info-dbus-service/io.edgehog.SystemInfo.conf
```diff
@@ -0,0 +1,33 @@
...
+  <!-- Only root can own the service -->
+  <policy user="root">
+    <allow own="io.edgehog.SystemInfo"/>
+    <allow send_destination="org.secomind.SystemInfo"/>
+    <allow send_interface="org.secomind.SystemInfo"/>
+  </policy>
+
+  <policy at_console="true">
+    <deny own="io.edgehog.SystemInfo"/>
+    <allow send_destination="org.secomind.SystemInfo"/>
```
harlem88, line 23: Let's use everywhere: `io.edgehog`

harlem88 (changes requested):

bettio (changes requested):
Here I have some comments about the logic behind this.

SystemInfo is meant to be the system identification, that is on a higher abstraction level in comparison to the hardware identification. Therefore I'm not sure that using any DMI or hardware identification is the right thing to do here.
The same kind of hardware might have different kind of systems, so the configuration should be taken elsewhere.
I think we might get rid of the DMI and d-bus stuff here, and use env or kernel cmdline as an easy way to inject the system part number and serial number.

src/telemetry/system_info.rs
```diff
@@ -0,0 +1,73 @@
...
+ */
+use crate::error::DeviceManagerError;
+use astarte_sdk::types::AstarteType;
+use procfs::cmdline;
+use std::collections::HashMap;
+use std::env;
+
+pub fn get_system_info() -> Result<HashMap<String, AstarteType>, DeviceManagerError> {
+    let cmdline_params = cmdline().unwrap();
+    let mut seral_number: Option<String> = env::var("SERIAL_NUMBER").ok();
```
bettio, line 28: SYSTEM_SERIAL_NUMBER

src/telemetry/system_info.rs
```diff
@@ -0,0 +1,73 @@
...
+    let mut part_number: Option<String> = env::var("PART_NUMBER").ok();
```
bettio, line 29: SYSTEM_PART_NUMBER

src/telemetry/system_info.rs
```diff
@@ -0,0 +1,73 @@
...
+    for param in cmdline_params.iter() {
+        if seral_number.is_none() && param.starts_with("serial_number") {
```
bettio, line 31: system_serial_number

src/telemetry/system_info.rs
```diff
@@ -0,0 +1,73 @@
...
+            let first_half = format!("{}=", "serial_number");
+            seral_number = Some(param.replace(&first_half, ""));
+        }
+        if part_number.is_none() && param.starts_with("part_number") {
```
bettio, line 35: system_part_number

bettio:
kernel documentation says:

> The kernel parses parameters from the kernel command line up to “–”; if it doesn’t recognize a parameter and it doesn’t contain a ‘.’, the parameter gets passed to init: parameters with ‘=’ go into init’s environment, others are passed as command line arguments to init. Everything after “–” is passed as an argument to init.

src/telemetry/system_info.rs
bettio: maybe also we should prefix all of the with `EDGHOG_`.

src/telemetry/system_info.rs
```diff
@@ -0,0 +1,73 @@
...
+use astarte_sdk::types::AstarteType;
+use procfs::cmdline;
+use std::collections::HashMap;
+use std::env;
+
+pub fn get_system_info() -> Result<HashMap<String, AstarteType>, DeviceManagerError> {
+    let cmdline_params = cmdline().unwrap();
+    let mut seral_number: Option<String> = env::var("SERIAL_NUMBER").ok();
+    let mut part_number: Option<String> = env::var("PART_NUMBER").ok();
+    for param in cmdline_params.iter() {
```
bettio, line 30: Do we really need to handle kernel command line?

bettio (changes requested):

harlem88 (approved):

bettio (approved):

PR #68, opened 2022-05-26
Add Repository for saving Runtime State

harlem88:
Repositories encapsulate the logic required to access data sources.
They centralize common data access functionality, providing better maintainability and decoupling the infrastructure or technology
used to access the data.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,6 +36,7 @@ pub mod error;
 mod ota_handler;
 mod power_management;
 mod rauc;
+mod repository;
 mod telemetry;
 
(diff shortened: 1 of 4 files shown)


bettio (commented):
Looks fairly good, however I'm worried that using a trait here right now might be a bit over engineering, so I would like to think about this or to ask comments

Annopaolo:
> Looks fairly good, however I'm worried that using a trait here right now might be a bit over engineering, so I would like to think about this or to ask comments

Do we plan to store state in different ways in the future? Then using a trait might be an elegant solution. However, if we're not planning to do so, it could be removed

PR #81, opened 2022-06-20
Remove single file storage

harlem88:
- Introduce a directory for data storage so that all data are confined to the same location and do not have files scattered throughout the system.
- Check configured path before device-runtime starts.
- Add some tests on get_credentials function.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,6 +18,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+use crate::repository::file_state_repository::FileStateRepository;
+use crate::repository::StateRepository;
 use astarte_sdk::builder::AstarteOptions;
 use astarte_sdk::types::AstarteType;
 use astarte_sdk::{registration, Aggregation, AstarteSdk};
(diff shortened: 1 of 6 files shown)


rbino (changes requested):
Additional note: this would break compatibility for existing devices with credentials saved already saved in the store file, they would have to be manually unregistered. Is this acceptable? @bettio

src/lib.rs
```diff
@@ -65,7 +65,15 @@ pub struct DeviceManager {
 impl DeviceManager {
     pub async fn new(opts: DeviceManagerOptions) -> Result<DeviceManager, DeviceManagerError> {
         let device_id: String = get_device_id(opts.device_id.clone()).await?;
-        let credential_secret: String = get_credentials_secret(&device_id, &opts).await?;
+        let credential_secret: String = get_credentials_secret(
```
Annopaolo, line 68: Using `credentials_secret` as variable name would be more consistent

src/lib.rs
```diff
@@ -65,7 +65,15 @@ pub struct DeviceManager {
...
+            &device_id,
+            &opts,
+            FileStateRepository::new(
+                opts.store_directory.clone(),
+                format!("credential_{}.json", device_id),
```
Annopaolo, line 73: See above, s/credential/credentials

Annopaolo (changes requested):
Just a minor comment

Annopaolo (approved):

rbino (approved):

bettio (approved):

PR #82, opened 2022-06-21
RAUC mark slot as good

harlem88:
Normally, the full system update chain is not complete before being sure that the newly installed system runs without any errors.
As the definition and detection of a successful operation is really  system-dependent, so we need to inform the bootloader that the slot selected by it was good.

### DIFF
--- a/src/ota/ota_handler.rs
+++ b/src/ota/ota_handler.rs
@@ -233,21 +233,53 @@ impl<'a> OTAHandler<'a> {
         if self.state_repository.exists() {
             info!("Found pending update");
             let state = self.state_repository.read()?;
+            let result_response = match self.do_pending_ota(&state).await {
+                Ok(()) => {
+                    info!("OTA successful");
+                    self.send_ota_response(sdk, &state.uuid, OTAStatus::Done)
+                        .await
+                }
+                Err(error) => {
+                    warn!("OTA failed, error -> {:?}", error);
+                    self.send_ota_response(sdk, &state.uuid, OTAStatus::Error(OTAError::Failed))
+                        .await
+                }
+            };
 
-            if state.slot != self.ota.boot_slot().await? {
-                info!("OTA successful");
-                self.send_ota_response(sdk, &state.uuid, OTAStatus::Done)
-                    .await?;
-            } else {
-                warn!("OTA failed");
-                self.send_ota_response(sdk, &state.uuid, OTAStatus::Error(OTAError::Failed))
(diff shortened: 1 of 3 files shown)


src/ota/ota_handler.rs
```diff
@@ -233,23 +233,54 @@ impl<'a> OTAHandler<'a> {
...
-                    .await?;
-            } else {
-                warn!("OTA failed");
-                self.send_ota_response(sdk, &state.uuid, OTAStatus::Error(OTAError::Failed))
-                    .await?;
+            if result_response.is_err() {
+                warn!(
+                    "Unable to publish OTA response -> {:?}",
+                    result_response.err()
+                );
```
rbino, line 253: Should this be retried (either today or in the future, leaving a TODO) on the device side or should the cloud side handle a missing answer from the device with a timeout? cc @bettio

bettio: Let's leave a TODO, but at least I would love avoiding timeouts.

bettio (changes requested):

src/ota/ota_handler.rs
```diff
@@ -473,6 +504,14 @@ mod tests {
 
         let mut state_mock = MockStateRepository::<PersistentState>::new();
         state_mock.expect_exists().returning(|| true);
+        ota.expect_get_primary()
+            .returning(|| Ok("rootfs.0".to_owned()));
```
bettio, line 508: are we hard-coding a slot name here? If so, I'm not a big fan of this.

harlem88: Is only in the test side.

src/ota/ota_handler.rs
```diff
@@ -233,21 +233,53 @@ impl<'a> OTAHandler<'a> {
...
+                }
+                Err(error) => {
+                    warn!("OTA failed, error -> {:?}", error);
+                    self.send_ota_response(sdk, &state.uuid, OTAStatus::Error(OTAError::Failed))
+                        .await
+                }
+            };
+
+            if result_response.is_err() {
+                warn!(
```
Annopaolo, line 250: Could this warn be moved in the `Err` branch of the match above?

harlem88: no, because is the result of `self.send_ota_response` returned by the match block.

Annopaolo (approved):

bettio (approved):

rbino (approved):

PR #115, opened 2022-07-19
Use consts::ARCH for reading the architecture of the CPU

harlem88:
- Use `std::env::consts::ARCH` for reading the architecture of the CPU
- Remove `nix` crate dependency


bettio:
Just one question: on 64 bit ARM CPUs what nix used to return?

PR #130, opened 2022-09-02
Implement io.edgehog.devicemanager.BatteryStatus

### LINKED ISSUE #16: Implement `io.edgehog.devicemanager.BatteryStatus` Interface in Device Runtime

FrancescoVaiani:
Add support to `io.edgehog.devicemanager.BatteryStatus` interface
with data from UPower collected via D-Bus.

Close #16 

Signed-off-by: Francesco Vaiani <francesco.vaiani@secomind.com>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -174,26 +174,7 @@ impl<T: Publisher + Clone + 'static> DeviceManager<T> {
         let publisher = self.publisher.clone();
         tokio::spawn(async move {
             while let Some(msg) = telemetry_rx.recv().await {
-                match msg.payload {
-                    TelemetryPayload::SystemStatus(data) => {
-                        let _ = publisher
-                            .send_object(
-                                "io.edgehog.devicemanager.SystemStatus",
-                                "/systemStatus",
-                                data,
-                            )
-                            .await;
-                    }
-                    TelemetryPayload::StorageUsage(data) => {
-                        let _ = publisher
-                            .send_object(
-                                "io.edgehog.devicemanager.StorageUsage",
-                                format!("/{}", msg.path).as_str(),
-                                data,
-                            )
-                            .await;
-                    }
-                };
+                Self::send_telemetry(&publisher, msg).await;
             }
         });
     }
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -174,26 +174,7 @@ impl<T: Publisher + Clone + 'static> DeviceManager<T> {
...
-                        let _ = publisher
-                            .send_object(
-                                "io.edgehog.devicemanager.StorageUsage",
-                                format!("/{}", msg.path).as_str(),
-                                data,
-                            )
-                            .await;
-                    }
-                };
+                send_telemetry(Box::new(&publisher), msg).await;
```
rbino, line 177: There's no need to allocate on the heap to pass the publisher reference:
- Move `send_telemetry` _inside_ the `DeviceManager` struct so you can use `T` and change its signature to:
```rust
    async fn send_telemetry(publisher: &T, msg: TelemetryMessage)
```
- At the callsite:
```rust
    Self::send_telemetry(&publisher, msg).await;
```

rbino (changes requested):

src/telemetry/battery_status.rs
```diff
@@ -0,0 +1,140 @@
...
+                level_percentage: device.percentage().await.unwrap(),
+                level_absolute_error,
+                status,
+            };
+
+            result.insert(device.serial().await?, battery);
+        }
+    }
+
+    Ok(result)
```
rbino, line 73: I thought of another approach for this that could reduce repetition without introducing back highly indented code: As you could have a `get_powet_devices() -> Result<Something>` function that chains the calls starting from `zbus::Connection::system` with the `?` operator and then the outer function just handles the `Ok/Err` case in a single place and either iterates on `power_devices` or returns an empty HashMap

rbino (changes requested):

harlem88 (approved):

rbino (approved):

bettio (approved):

PR #136, opened 2022-09-14
Add support to `CellularConnectionProperties`

### LINKED ISSUE #17: Implement `io.edgehog.devicemanager.CellularConnectionProperties` Interface in Device Runtime

FrancescoVaiani:
- Implement the Cellular Properties D-Bus demo service
- Add support to the `Modems` D-Bus service to retrieve the data for `io.edgehog.devicemanager.CellularConnectionProperties`

Close #17


src/telemetry/cellular_modem.rs
```diff
@@ -0,0 +1,103 @@
...
+    if let Ok(connection) = zbus::Connection::session().await {
+        if let Ok(proxy) = CellularModemsProxy::new(&connection).await {
+            if let Ok(available_modems) = proxy.list().await {
+                for modem_id in available_modems {
+                    let modem: ModemProperties = proxy.get(modem_id.clone()).await?;
+                    ret.extend(get_modem_properties(modem_id, modem));
+                }
+            }
+        }
+    }
```
rbino, line 56: As in the other PR, let's try to keep (at least in new code) an early return approach vs the heavily nested approach here

rbino: As in the other PR, you could have a `get_available_modems() -> Result<Something>` function that chains the calls starting from `zbus::Connection::session` with the `?` operator and then the outer function just handles the `Ok/Err` case in a single place and either iterates on `available_modems` or returns an empty HashMap

rbino (changes requested):

rbino (approved):

joshuachp:
Superseded by https://github.com/edgehog-device-manager/edgehog-device-runtime/pull/402

PR #168, opened 2022-10-14
Remove update file after installation

FrancescoVaiani:
The OTA update file is left on the disk after the OTA update even if it is not needed anymore.
In addition to removing the file after the installation, an additional check is performed on the eventually existing file before downloading.

Signed-off-by: Francesco Vaiani <francesco.vaiani@secomind.com>

### DIFF
--- a/src/ota/ota_handler.rs
+++ b/src/ota/ota_handler.rs
@@ -20,6 +20,7 @@
 
 use astarte_device_sdk::AstarteAggregate;
 use std::collections::HashMap;
+use std::path::PathBuf;
 
 use astarte_device_sdk::types::AstarteType;
 use log::{debug, error, info, warn};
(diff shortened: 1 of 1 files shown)


FrancescoVaiani:
⚠PR rebased on #88 for improved compatibility. Please merge after the SDK update.

src/ota/ota_handler.rs
```diff
@@ -104,6 +105,10 @@ impl<'a> OTAHandler<'a> {
         self.ota.last_error().await
     }
 
+    pub fn get_update_file_path(&self) -> PathBuf {
```
harlem88, line 108: Why declare this function pub?

harlem88 (changes requested):

PR #206, opened 2023-03-14
Add Astarte Message Hub

### LINKED ISSUE #169: Add support to Astarte Message Hub
- [x] Make it configurable, the device runtime should be support both connection ways ( Astarte SDK,  Astarte Message Hub )
- [x] The codebase should be the same, I think should be added a new impl of `Publisher`

harlem88:
~~In the AstarteMessageHubNode struct I used a Mutex instead of RwLock, because the tonic::transport::Streaming is !Sync so it can be shared between threads only using a Mutex (this is an interesting difference between the two).~~

- Separate Astarte Subscriber from Publisher.
- Add Protoc install to CI.

closes #169

### DIFF
--- a/src/data/astarte_device_sdk_lib.rs
+++ b/src/data/astarte_device_sdk_lib.rs
@@ -18,26 +18,36 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-use astarte_device_sdk::options::AstarteOptions;
+use astarte_device_sdk::options::{AstarteOptions, AstarteOptionsError};
 use astarte_device_sdk::types::AstarteType;
 use astarte_device_sdk::{
     registration, AstarteAggregate, AstarteDeviceDataEvent, AstarteDeviceSdk, AstarteError,
 };
 use async_trait::async_trait;
+use serde::Deserialize;
 
-use crate::data::Publisher;
+use crate::data::{Publisher, Subscriber};
 use crate::error::DeviceManagerError;
 use crate::repository::file_state_repository::FileStateRepository;
 use crate::repository::StateRepository;
 use crate::{get_hardware_id_from_dbus, DeviceManagerOptions};
 
+#[derive(Debug, Deserialize, Clone)]
+pub struct AstarteDeviceSdkConfigOptions {
+    pub realm: String,
+    pub device_id: Option<String>,
+    pub credentials_secret: Option<String>,
+    pub pairing_url: String,
+    pub pairing_token: Option<String>,
+}
+
 #[derive(Clone)]
-pub struct Astarte {
+pub struct AstarteDeviceSdkLib {
     pub device_sdk: AstarteDeviceSdk,
 }
 
 #[async_trait]
-impl Publisher for Astarte {
+impl Publisher for AstarteDeviceSdkLib {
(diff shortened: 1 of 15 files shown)


joshuachp (changes requested):
We are discarding a lot of the errors, this could make it hard to debug or solve issues later. It could be better if at least we print the error message with a format, or even better put the error as a #[source] so it can be then reported

src/data/astarte_message_hub_node/node.rs
```diff
@@ -0,0 +1,186 @@
...
+
+impl AstarteMessageHubNode {
+    pub async fn new(options: AstarteMessageHubOptions) -> Result<Self, DeviceManagerError> {
+        use proto_message_hub::message_hub_client::MessageHubClient;
+        let mut message_hub_client = MessageHubClient::connect(options.endpoint).await?;
+
+        let interface_json = read_interfaces_from_directory(&options.interfaces_directory)?;
+
+        let node = proto_message_hub::Node {
+            uuid: "d72a6187-7cf1-44cc-87e8-e991936166db".to_string(),
```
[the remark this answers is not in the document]
harlem88: It is the node id that identifies the Device runtime in each system/deployment, perhaps we should declare it as const.

src/data/astarte.rs
```diff
@@ -18,19 +18,29 @@
  * SPDX-License-Identifier: Apache-2.0
```
sorru94, line 18: This file is the counterpart of `astarte_message_hub_node.rs` that uses the SDK right?
It might be clearer to rename it to something that indicates that. Like `astarte_sdk_device`.

src/data/astarte_message_hub_node.rs
```diff
@@ -0,0 +1,734 @@
...
+    endpoint: String,
+}
+
+/// Struct containing the configuration options for the Astarte message hub node.
+pub struct AstarteMessageHubNodeOptions {
+    /// The Endpoint of the Astarte Message Hub
+    endpoint: String,
+    /// Directory containing the Astarte interfaces.
+    interfaces_directory: String,
+}
```
sorru94, line 51: These two structs seem redundant.
The only reason for the existence of `AstarteMessageHubNodeOptions` is to provide the `new` function with the `interfaces_directory` right?
Can't we reach the same goal by passing the `interfaces_directory` as a separate argument?

src/data/astarte_device_sdk_lib.rs
```diff
@@ -18,19 +18,29 @@
...
+pub struct AstarteDeviceSdkConfigOptions {
+    pub realm: String,
+    pub device_id: Option<String>,
+    pub credentials_secret: Option<String>,
+    pub pairing_url: String,
+    pub pairing_token: Option<String>,
+}
+
 #[derive(Clone)]
 pub struct Astarte {
```
sorru94, line 45: I would take this chance to rename this struct in something more indicative of its use.
Something along the lines of: `AstarteDeviceSdkInstance`, `AstarteDevice`, `AstarteDeviceSdkLib`.

sorru94 (approved):

bettio (approved):

PR #236, opened 2023-06-26
Add release branch to CI

harlem88:


joshuachp (approved):
SGTM, we should add it to other repos too

bettio (approved):

PR #242, opened 2023-07-12
chore(deps): pin dependencies to work with MSRV

joshuachp:
Since it's hard to work with the MSRV this PR is to pin the transitive dependencies to the one working with the MSRV. This makes it easier to add, remove or update dependencies without breaking the build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ target/
 
 .idea/
 /telemetry.json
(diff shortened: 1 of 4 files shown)


joshuachp:
The `value-bag` dependency will fail to compile on future versions of rust

harlem88 (approved):

PR #252, opened 2023-07-24
tests: reduce the test time from >30s to <1s

joshuachp:
Remove the delays and sleep in the tests, optimize the channels into a stream. Fix flaky tests when run locally, use temp dir for ota.bin. Use tokio process and uitls to not block the main thread.

Advance time manually during tests.

### DIFF
--- a/src/ota/ota_handle.rs
+++ b/src/ota/ota_handle.rs
@@ -457,13 +457,14 @@ where
 
         info!("Rebooting in 5 seconds");
 
-        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+        if cfg!(not(test)) {
+            tokio::time::sleep(std::time::Duration::from_secs(5)).await;
 
-        #[cfg(not(test))]
-        if let Err(error) = crate::power_management::reboot() {
-            let message = "Unable to run reboot command";
-            error!("{message} : {error}");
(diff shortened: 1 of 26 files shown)


rgallor (approved):

rgallor (approved):

src/ota/ota_handle.rs
```diff
@@ -457,13 +457,14 @@ where
 
         info!("Rebooting in 5 seconds");
 
-        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+        if cfg!(not(test)) {
+            tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
harlem88: `cfg!`, unlike `#[cfg]`, does not remove any code and only evaluates to true or false. [std::cfg](https://doc.rust-lang.org/std/macro.cfg.html)
Keep the version with `#[cfg]`
``` Rust
#[cfg(not(test))]{
   tokio::time::sleep(std::time::Duration::from_secs(5)).await;
   if let Err(error) = crate::power_management::reboot() {
...
}

harlem88 (changes requested):

harlem88 (approved):

sorru94 (approved):

PR #287, opened 2023-10-12
feat(http): implement http over websocket

### LINKED ISSUE #275: [Forwarder]: Implement HTTP protocol over websocket

rgallor:
To enhance code comprehension, it is advisable to follow this order when inspecting the code files:
1. `edgehog-device-runtime/src/forwarder.rs`: This module is responsible for initiating a new forwarder instance upon receiving an Astarte event. A forwarder task is created, which, in turn, generates a `ConnectionsManager` (see point 3) responsible for handling all connection-related tasks.
2. `edgehog-device-runtime/src/lib.rs`: In this file, the `DeviceManager` now includes a new field called `forwarder_handler`, which utilizes the `forwarder.rs` module when an Astarte event pertaining to the `RemoteTerminalRequest` interface is received and converted into a `ConnectionInfo` struct (defined in `edgehog-device-runtime-forwarder/src/astarte.rs`).
3. `edgehog-device-runtime-forwarder/src/connections_manager.rs`: This file defines the `ConnectionsManager` struct, which is accountable for establishing a WebSocket connection between the device and Edgehog. Additionally, it functions as a multiplexer to forward messages between Edgehog and internal device connections (e.g., with TTYD). When it receives a WebSocket frame from Edgehog, it first extracts the frame and, if it is binary, decodes the `protobuf` data into an internal Rust representation defined in `edgehog-device-runtime-forwarder/src/messages.rs`. It also handles HTTP requests (WebSocket's requests are still pending). In case a device connection (e.g., with TTYD) (see point 4) sends data to the manager, it encodes the message into a WebSocket frame and forwards it to Edgehog. The manager is also responsible for reconnection operations in case a WebSocket error occurs during communication. The `Connections` structure serves as a collection of connections and maintains a channel writing handle necessary for each device connection to forward messages to the manager.
4. `edgehog-device-runtime-forwarder/src/connection.rs`: This file defines the `Connection` and `ConnectionHandle` structs, which are responsible for creating a new connection and spawning the Tokio task responsible for managing it. Currently, a connection task sends an HTTP request (e.g., to TTYD), waits for a response, encodes it into a protobuf message, and forwards it to the `ConnectionsManager`.

Close #275


edgehog-device-runtime-forwarder/src/messages.rs
```diff
@@ -0,0 +1,468 @@
...
+    Http(#[from] http::Error),
+    /// Error while parsing Headers.
+    ParseHeaders(#[from] http::header::ToStrError),
+    /// Invalid port number
+    InvalidPortNumber(#[from] TryFromIntError),
+}
+
+/// Requests Id.
+#[derive(Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub struct Id(Vec<u8>);
```
harlem88, line 71: Is it really necessary to define this struct and its methods and not use only `Vec<u8>`?

harlem88 (changes requested):

edgehog-device-runtime-forwarder/src/messages.rs
rgallor: I think it is better to implement it. It's necessary to implement Debug with `hex` output and it's a clearer name (for instance, instead of passing `Vec<u8>` to all functions). Also it is easier to refactor in the future.

edgehog-device-runtime-forwarder/src/messages.rs
```diff
@@ -0,0 +1,468 @@
...
+        ProtoMessage::try_from(msg)
+    }
+}
+
+/// Supported protocols.
+#[derive(Debug)]
+pub(crate) enum Protocol {
+    Http(Http),
+    WebSocket(WebSocket),
+}
```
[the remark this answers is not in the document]
rgallor: For instance, port number in protobuf is an `u32`. Instead with an internal representation I can validate each field using an `u16`. I can also implements methods on this repo's structs instead on the proto's one. This way, it's easier to change something when the proto files are changed (for example, the `request_id`).

edgehog-device-runtime-forwarder/src/connections.rs
```diff
@@ -0,0 +1,362 @@
...
+        Ok(())
+    }
+
+    /// Close all the connections the device has established (e.g., with TTYD).
+    #[instrument(skip_all)]
+    pub(crate) fn disconnect(&mut self) {
+        info!("closing all the connections");
+        self.connections.disconnect();
+    }
+}
```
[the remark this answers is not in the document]
rgallor: I kept this struct and the respective methods just because everything in this file was related to the connections management. Also because the other structures in `connections.rs` are the errors and `Connections` (which are slightly small). Anyway, it seems also reasonable to move it into a separate file. What do you think?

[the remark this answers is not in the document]
harlem88: Yes, I would prefer to move it to a separate file.

edgehog-device-runtime-forwarder/src/messages.rs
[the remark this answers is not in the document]
harlem88: Ok in this case I agree with you to not bind with code generated by proto.

edgehog-device-runtime-forwarder/src/connections.rs
```diff
@@ -0,0 +1,362 @@
...
+    #[instrument]
+    pub async fn connect(url: Url) -> Result<Self, Error> {
+        // TODO: check if, when a wrong URL is passed, it will endlessly try to connect
+
+        let ws_stream = Self::ws_connect(&url).await?;
+
+        // this channel is used by tasks associated to each connection to communicate new
+        // information available on a given websocket between the device and TTYD.
+        // it is also used to forward the incoming data from TTYD to the device.
+        let (tx_ws, rx_ws) = unbounded_channel();
```
joshuachp, line 94: Set a bound to the maximum number of connection's messages

src/forwarder.rs
```diff
@@ -0,0 +1,120 @@
...
+impl Eq for Key {}
+
+impl Hash for Key {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.host.hash(state);
+        self.port.hash(state);
+    }
+}
+
+/// Device forwarder.
```
joshuachp, line 62: Add a comment on why this differs from Connections

edgehog-device-runtime-forwarder/src/connections.rs
```diff
@@ -0,0 +1,362 @@
...
+                error!("Http response should not be sent by the bridge");
+                return Err(Error::WrongMessage);
+            }
+            ProtoProtocol::WebSocket(_ws) => {
+                error!("WebSocket messages are not supported yet");
+                return Err(Error::WrongMessage);
+            }
+        };
+
+        self.connections.add_http(http_req)?;
```
joshuachp, line 274: rename this into handle_http and move it in the respective match arm

edgehog-device-runtime-forwarder/src/connections.rs
```diff
@@ -0,0 +1,362 @@
...
+    /// Create a new [`Connection`] in case a new HTTP request is received.
+    #[instrument(skip(self, http_req))]
+    pub(crate) fn add_http(&mut self, http_req: HttpRequest) -> Result<(), Error> {
+        let id = http_req.request_id.clone();
+
+        // check if there exist a connection with that id
+        let Entry::Vacant(entry) = self.connections.entry(id.clone()) else {
+            error!("entry already occupied");
+            return Err(Error::IdAlreadyUsed(id));
+        };
```
joshuachp, line 335: Intead of on every message to check if all the entries are finished. Create a method like `connection` that will return an entry only if is not finished and will remove the finished one.

joshuachp: Also move the remove_terminated_connections into a function `add_connection` that will cleanup all the hashmap.

joshuachp (changes requested):
LGTM other than the changes requested. I would also add a feature for the forwarded and feature gate the calls in the edgehog_device_runtime

edgehog-device-runtime-forwarder/src/connections.rs
rgallor: What do you think a good number of messages inside the channel could be?

joshuachp: In the astarte_device_sdk we use 50 as an arbitrary value

joshuachp: Another strategy could be using the crate `num_cpu`;

edgehog-device-runtime-forwarder/src/connections.rs
```diff
@@ -0,0 +1,113 @@
...
+    pub(crate) fn remove_terminated(&mut self) {
+        self.connections
+            .retain(|_k, con_handle| !con_handle.is_finished());
+    }
+
+    /// Terminate all connections, both active and non.
+    pub(crate) fn disconnect(&mut self) {
+        self.connections
+            .values_mut()
+            .for_each(|con| con.abort_handle().abort());
```
joshuachp, line 109: You can directly call abort on the `JoinHandle`

edgehog-device-runtime-forwarder/src/connections_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+        }
+
+        Ok(())
+    }
+
+    /// Handle a single connection event.
+    #[instrument(skip_all)]
+    pub(crate) async fn event_loop(&mut self) -> Result<ControlFlow<()>, TungError> {
+        let event = self
+            .select_ws_event(Duration::from_secs(PING_TIMEOUT))
```
joshuachp, line 137: Make the timeout `Duration` a const since from_secs is const (since 1.32) and pass it directly to tokio::time::timeout (in the select) instead of constructing it each iteration and passing it to the function, since duration is Copy

joshuachp (changes requested):

edgehog-device-runtime-forwarder/src/messages.rs
```diff
@@ -0,0 +1,468 @@
...
+    /// Create a close frame.
+    pub(crate) fn close(code: u16, reason: Option<String>) -> Self {
+        Self::Close { code, reason }
+    }
+}
+
+/// Convert a [`HeaderMap`] containing all HTTP headers into a [`HashMap`].
+pub(crate) fn headermap_to_hashmap<'a, I>(headers: I) -> Result<HashMap<String, String>, ProtoError>
+where
+    I: Iterator<Item = (&'a HeaderName, &'a HeaderValue)>,
```
joshuachp, line 283: I believe this could be better if the trait bound is on IntoIterator instead

src/forwarder.rs
```diff
@@ -0,0 +1,124 @@
...
+            Ok(cinfo) => cinfo,
+            // error while retrieving the connection information from the Astarte data
+            Err(err) => {
+                error!("{err}");
+                return;
+            }
+        };
+
+        // check if the remote terminal task is already running. if not, spawn a new task and add it to the collection
+        self.get_running(cinfo.clone()).or_insert(
```
joshuachp, line 96: This can be `or_insert_with_key` so we don't need to clone the key. Maybe you need to `impl TryFom<&ConnectionInfo> for Url` instead of the owned value one.

src/forwarder.rs
```diff
@@ -0,0 +1,124 @@
...
+            // spawn a new task responsible for handling the remote terminal operations
+            Self::spawn_task(cinfo),
+        );
+    }
```
joshuachp, line 100:
```suggestion
    }

```

src/forwarder.rs
```diff
@@ -0,0 +1,124 @@
...
+        // remove all finished tasks
+        self.tasks.retain(|_, jh| !jh.is_finished());
+
+        self.tasks.entry(Key(cinfo))
+    }
+
+    /// Task handling a device forwarder instance.
+    fn spawn_task(cinfo: ConnectionInfo) -> JoinHandle<()> {
+        tokio::spawn(async move {
+            let edgehog_url = Url::try_from(cinfo).unwrap();
```
joshuachp, line 111: Do not unwrap here and use the error! macro

joshuachp (changes requested):

src/forwarder.rs
```diff
@@ -0,0 +1,137 @@
...
+
+use astarte_device_sdk::{Aggregation, AstarteDeviceDataEvent};
+use edgehog_forwarder::astarte::{retrieve_connection_info, ConnectionInfo};
+use edgehog_forwarder::connections_manager::ConnectionsManager;
+use log::error;
+use reqwest::Url;
+use tokio::task::JoinHandle;
+
+#[derive(Debug)]
+struct Key(ConnectionInfo);
```
harlem88, line 38: In this case you cannot use ConnectionInfo directly?  and Add the `impl traits` directly to ConnectionInfo? you often use connectionInfo so I think `Key` struct it's extra.

harlem88 (changes requested):

harlem88 (changes requested):

edgehog-device-runtime-forwarder/src/connections_manager.rs
```diff
@@ -0,0 +1,292 @@
...
+                    Ok(proto_msg) => self.handle_proto_msg(proto_msg),
+                    Err(err) => {
+                        error!("failed to decode protobuf message due to {err:?}");
+                        Ok(())
+                    }
+                };
+
+                if let Err(err) = res {
+                    error!("failed to handle protobuf message due to {err:?}");
+                }
```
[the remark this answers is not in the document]
rgallor: Done

src/forwarder.rs
```diff
@@ -71,21 +103,57 @@ impl Hash for Key {
...
 /// the connection information and responsible for providing forwarder functionalities. For
 /// instance, a task could open a remote terminal between the device and a certain host.
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct Forwarder {
+    tx_state: Sender<SessionState>,
     tasks: HashMap<Key, JoinHandle<()>>,
 }
 
+const CHANNEL_STATE_SIZE: usize = 50;
```
harlem88, line 112: Move this definition on top, after the `use` but before the structs declaration.

harlem88 (changes requested):

edgehog-device-runtime-forwarder/src/connections.rs
harlem88: Use just `num_cpu` could be provide a strange behavior on board with single or double CPU, maybe we could add a multiplier by `num_cpu`, for instance:
`12 * num_cpu`

edgehog-device-runtime-forwarder/src/astarte.rs
```diff
@@ -0,0 +1,211 @@
...
+        Ok(hm)
+    }
+}
+
+impl TryFrom<&SessionInfo> for Url {
+    type Error = AstarteError;
+
+    fn try_from(value: &SessionInfo) -> Result<Self, Self::Error> {
+        if value.session_token.is_empty() {
+            return Err(AstarteError::MissingUrlInfo("session token"));
```
harlem88, line 74:
```suggestion
            return Err(AstarteError::MissingUrlInfo("Missing session token"));
```

edgehog-device-runtime-forwarder/src/connections_manager.rs
```diff
@@ -0,0 +1,304 @@
...
+    #[instrument]
+    pub async fn connect(url: Url) -> Result<Self, Error> {
+        // TODO: check if, when a wrong URL is passed, it will endlessly try to connect
+
+        let ws_stream = Self::ws_connect(&url).await?;
+
+        // this channel is used by tasks associated to each connection to communicate new
+        // information available on a given websocket between the device and TTYD.
+        // it is also used to forward the incoming data from TTYD to the device.
+        let (tx_ws, rx_ws) = channel(CHANNEL_SIZE);
```
harlem88, line 85: So in this case the `CHANNEL_SIZE` is the number of open forwarding connections?

src/forwarder.rs
```diff
@@ -0,0 +1,387 @@
...
+}
+
+impl Forwarder {
+    /// Initialize the forwarder instance, spawning also a task responsible for sending a property
+    /// necessary to update the device session state.
+    pub fn init<P>(publisher: P) -> Self
+    where
+        P: Publisher + 'static,
+    {
+        let (tx_state, rx_state) = channel::<SessionState>(CHANNEL_STATE_SIZE);
```
harlem88, line 146: Perhaps again `50` is too much, you are considering that at the same time 50 forwarders are trying to update their status?

harlem88 (changes requested):

edgehog-device-runtime-forwarder/src/connections_manager.rs
rgallor: Within a session established between the device and the bridge the `ConnectionsManager` uses this channel to receive data from the internal device connections (such as those with TTYD). A single forwarding connection may send more data on the channel, thus it is unlikely that CHANNEL_SIZE is effectively the same as the number of connections.

src/forwarder.rs
rgallor: Yes maybe this number is a little bit high, since it is improbable to have 50 sessions opened at the same time and requesting a change of state. What do you think it could be a good estimate of the number of sessions on a given device? (Imo 5 could be enough)

harlem88: Yes I think 5 is reasonable number for starting.

Cargo.toml
```diff
@@ -66,19 +68,25 @@ tokio = { workspace = true, features = ["test-util"] }
...
 astarte-device-sdk = "0.5.1"
 astarte-message-hub = "0.5.2"
 async-trait = "0.1.72"
+backoff = "0.4.0"
 base64 = "0.13.1"
 bollard = "0.15.0"
 bytes = "1.5.0"
 clap = "4.3.19"
 displaydoc = "0.2.4"
+edgehog-device-forwarder-proto = { git = "https://github.com/edgehog-device-manager/edgehog-device-forwarder-proto.git" }
```
harlem88, line 84: Change to `alpha` release published on `crates.io`.
```suggestion
edgehog-device-forwarder-proto = "0.1.0-alpha.0"
```

harlem88 (changes requested):

Cargo.toml
rgallor: Thanks for the reminder. I already changed it but it seems i undid the changes somehow.

atsetilam (approved):

Cargo.toml
rgallor: Done

edgehog-device-runtime-forwarder/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+The end user triggers a remote terminal request on the Edgehog Device Manager interface. Edgehog is responsible for
+sending the request to Astarte, which then forwards it to the Device Runtime, handling device operations.
+The Device Runtime retrieves session information from the remote terminal request and communicates it to the Forwarder,
+a module capable of managing connections with the device.
+Using the received information, the Forwarder establishes a WebSocket connection with Edgehog. On top of this
+connection, other types of connections (HTTP, WebSocket, etc.) can be established. At this point of the implementation, 
+the library only supports HTTP connections. For instance, the Service receiving HTTP requests could be an HTTP server 
+providing access to some device resources. In this scenario, Edgehog makes one or more HTTP requests to the Forwarder, 
+which forwards them to the Service. Then the service replies with one or more HTTP responses, that will be forwarded
```
atsetilam, line 45: **S**ervice

edgehog-device-runtime-forwarder/README.md
```diff
@@ -0,0 +1,91 @@
...
+The end user triggers a remote terminal request on the Edgehog Device Manager interface. Edgehog is responsible for
+sending the request to Astarte, which then forwards it to the Device Runtime, handling device operations.
+The Device Runtime retrieves session information from the remote terminal request and communicates it to the Forwarder,
+a module capable of managing connections with the device.
+Using the received information, the Forwarder establishes a WebSocket connection with Edgehog. On top of this
+connection, other types of connections (HTTP, WebSocket, etc.) can be established. At this point of the implementation, 
+the library only supports HTTP connections. For instance, the Service receiving HTTP requests could be an HTTP server 
+providing access to some device resources. In this scenario, Edgehog makes one or more HTTP requests to the Forwarder, 
+which forwards them to the Service. Then the service replies with one or more HTTP responses, that will be forwarded 
+to Edgehog
```
atsetilam, line 46: full stop after _Edgehog_**.**

atsetilam (commented):
Nitpicking and hairsplitting

edgehog-device-runtime-forwarder/README.md
```diff
@@ -0,0 +1,91 @@
+<!---
+  Copyright 2023 SECO Mind Srl
```
harlem88, line 2:
```suggestion
  Copyright 2024 SECO Mind Srl
```

edgehog-device-runtime-forwarder/src/astarte.rs
```diff
@@ -0,0 +1,211 @@
+// This file is part of Edgehog.
+//
+// Copyright 2023 SECO Mind Srl
```
harlem88, line 3:
```suggestion
// Copyright 2024 SECO Mind Srl
```

edgehog-device-runtime-forwarder/src/collection.rs
```diff
@@ -0,0 +1,157 @@
+// Copyright 2023 SECO Mind Srl
```
harlem88, line 1:
```suggestion
// Copyright 2024 SECO Mind Srl
```

edgehog-device-runtime-forwarder/src/connection.rs
```diff
@@ -0,0 +1,132 @@
+// This file is part of Edgehog.
+//
+// Copyright 2023 SECO Mind Srl
```
harlem88, line 3:
```suggestion
// Copyright 2024 SECO Mind Srl
```

edgehog-device-runtime-forwarder/src/connections_manager.rs
```diff
@@ -0,0 +1,323 @@
+// Copyright 2023 SECO Mind Srl
```
harlem88, line 1:
```suggestion
// Copyright 2024 SECO Mind Srl
```

edgehog-device-runtime-forwarder/src/messages.rs
```diff
@@ -0,0 +1,716 @@
+// This file is part of Edgehog.
+//
+// Copyright 2023 SECO Mind Srl
```
harlem88, line 3:
```suggestion
// Copyright 2024 SECO Mind Srl
```

edgehog-device-runtime-forwarder/tests/http_test.rs
```diff
@@ -0,0 +1,173 @@
+// Copyright 2023 SECO Mind Srl
```
harlem88, line 1:
```suggestion
// Copyright 2024 SECO Mind Srl
```

src/forwarder.rs
```diff
@@ -0,0 +1,387 @@
+/*
+ * This file is part of Edgehog.
+ *
+ * Copyright 2023 SECO Mind Srl
```
harlem88, line 4:
```suggestion
 * Copyright 2024 SECO Mind Srl
```

harlem88 (changes requested):
LGTM, only few things on license year

harlem88 (approved):

.github/workflows/codecov.yaml
```diff
@@ -58,7 +58,7 @@ jobs:
         uses: actions-rs/tarpaulin@v0.1
         with:
           version: '0.20.1'
-          args: '--ignore-tests --lib --all-features -- --test-threads 1'
+          args: '--ignore-tests --packages edgehog-device-runtime edgehog-device-runtime-docker edgehog-device-runtime-forwarder --all-features -- --test-threads 1'
```
joshuachp, line 61: Since `--lib` has been removed, add `--exclude-files` to ignore the main.rs

Cargo.toml
```diff
@@ -34,6 +34,8 @@ bytes = { workspace = true }
 clap = { workspace = true, features = ["derive"] }
 displaydoc = { workspace = true }
 env_logger = { workspace = true }
+edgehog-device-forwarder-proto = { workspace = true, optional = true }
```
joshuachp, line 37: Is this needed in the `edgehog-device-runtime` package?

Cargo.toml
```diff
@@ -73,22 +75,28 @@ tokio-stream = { workspace = true, features = ["net"] }
 [features]
 message-hub = ["astarte-device-sdk/message-hub"]
 systemd = ["dep:systemd"]
+forwarder = ["dep:edgehog-forwarder", "dep:edgehog-device-forwarder-proto"]
```
joshuachp, line 78: Clean up also here for the above comment

Cargo.toml
```diff
@@ -103,11 +111,14 @@ tempdir = "0.3.7"
 thiserror = "1.0.56"
 tokio = "1.35.1"
 tokio-stream = "0.1.14"
+tokio-tungstenite = "0.20.1"
 tokio-util = "0.7.10"
 toml = "0.8.2"
 tonic = "0.10.2"
 tracing = "0.1.40"
+tungstenite = "0.20.1"
```
joshuachp, line 119: This can be removed, since we use `tokio-tungsdenite`.

edgehog-device-runtime-forwarder/Cargo.toml
```diff
@@ -26,3 +26,22 @@ rust-version = { workspace = true }
...
+edgehog-device-forwarder-proto = { workspace = true }
+futures = { workspace = true }
+hex = { workspace = true }
+http = { workspace = true }
+reqwest = { workspace = true }
+thiserror = { workspace = true }
+tokio = { workspace = true }
+tokio-tungstenite = { workspace = true }
+tracing = { workspace = true, features = ["log"] }
+tungstenite = { workspace = true }
```
joshuachp, line 41: This can also be removed

joshuachp (changes requested):

Cargo.toml
rgallor: No it isn't, thanks for making me notice it

edgehog-device-runtime-forwarder/Cargo.toml
rgallor: Apart from `tungstenite`, the other deps are necessary

edgehog-device-runtime-forwarder/src/connection.rs
```diff
@@ -0,0 +1,116 @@
...
+    pub(crate) fn spawn(self) -> ConnectionHandle {
+        // spawn a task responsible for notifying when new data is available
+        let handle = tokio::spawn(async move {
+            // the span in used to know from which request a possible error is generated
+            let span = span!(Level::DEBUG, "spawn", id = %self.id);
+            let _enter = span.enter();
+
+            if let Err(err) = self.task().await {
+                error!("connection task failed with error {err:?}");
+            }
```
joshuachp, line 86: This should be in a function with instrument, to be thread local compatible (impl detail for tokio and tracing)

edgehog-device-runtime-forwarder/src/connections_manager.rs
```diff
@@ -0,0 +1,322 @@
...
+                        )?)))
+                    }
+                    Err(err) => {
+                        debug!("try reconnecting with backoff after tungstenite error: {err}");
+                        Err(BackoffError::Transient {
+                            err: Error::WebSocket(err),
+                            retry_after: None,
+                        })
+                    }
+                }
```
joshuachp, line 120:
```suggestion
                connect_async(url).await
                    .map_err(|err| match err {
                        TungError::Http(http_res) if http_res.status().is_client_error() => {
                            error!(
                                "received HTTP client error ({}), stopping backoff",
                                http_res.status()
                            );
                            let token = get_token(url)?;
                            BackoffError::Permanent(Error::TokenAlreadyUsed())
                        },
                        err => {
                            debug!("try reconnecting with backoff after tungstenite error: {err}");
                            BackoffError::Transient {
                                err: Error::WebSocket(err),
                                retry_after: None,
                            }
                        }
                    })

```

joshuachp: Add ca comment on the TungError::Http to explain why we terminate the backoff

src/forwarder.rs
```diff
@@ -0,0 +1,387 @@
...
+
+        // check if the remote terminal task is already running. if not, spawn a new task and add it to the collection
+        let tx_state = self.tx_state.clone();
+        let session_token = cinfo.session_token.clone();
+        self.get_running(cinfo).or_insert_with(||
+            // spawn a new task responsible for handling the remote terminal operations
+            Self::spawn_task(bridge_url, session_token, tx_state));
+    }
+
+    fn retrieve_astarte_data(
```
joshuachp, line 216: this function should return an error since the interface that was passed is incorrect

joshuachp: Like retrieve connection info

src/forwarder.rs
```diff
@@ -0,0 +1,387 @@
...
+
+        match ConnectionsManager::connect(bridge_url.clone()).await {
+            Ok(mut con_manager) => {
+                // update the session state to "connected"
+                if let Err(err) = tx_state
+                    .send(SessionState::connected(session_token.clone()))
+                    .await
+                {
+                    error!("failed to change session state to connected, {err}");
+                }
```
joshuachp, line 271: This should exit since the channel has been close

src/forwarder.rs
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    /// Spawn the task responsible for handling a device forwarder instance.
+    fn spawn_task(
+        bridge_url: Url,
+        session_token: String,
+        tx_state: Sender<SessionState>,
+    ) -> JoinHandle<()> {
+        tokio::spawn(Self::handle_session(bridge_url, session_token, tx_state))
+    }
```
joshuachp, line 247: This can probabli be inlined

src/forwarder.rs
```diff
@@ -0,0 +1,394 @@
...
+                    "io.edgehog.devicemanager.ForwarderSessionState",
+                    &ipath,
+                    idata,
+                )
+                .await
+            {
+                error!("publisher send error, {err}");
+            }
+        }
+    }
```
joshuachp, line 182: This can be handle in the connection task

src/forwarder.rs
```diff
@@ -0,0 +1,394 @@
...
+            Err(err) => {
+                // unset the session state, meaning that the device correctly disconnected itself
+                if let Err(err) = tx_state
+                    .send(SessionState::disconnected(session_token))
+                    .await
+                {
+                    error!("failed to change session state to disconnected, {err}");
+                }
+                error!("failed to connect, {err}")
+            }
```
joshuachp, line 295: Can we use a let Ok() ... else here to remove a lavel of indentation

joshuachp (changes requested):

src/forwarder.rs
rgallor: by using `cargo fmt` it still remains the same

src/forwarder.rs
rgallor: by spawning it in the `init` function the channel is being defined only, independently from the opened session. 
Btw, where do you mean exactly?

src/forwarder.rs
```diff
@@ -0,0 +1,327 @@
...
+        if let Err(err) = publisher
+            .send(
+                "io.edgehog.devicemanager.ForwarderSessionState",
+                &ipath,
+                idata,
+            )
+            .await
+        {
+            error!("publisher send error, {err}");
+        }
```
joshuachp, line 147: I would make this an hard error, disconnecting the forwarder

joshuachp (changes requested):

joshuachp (approved):

harlem88 (approved):

harlem88 (approved):

joshuachp (approved):

PR #291, opened 2023-10-24
feat(docker): create image req

### LINKED ISSUE #277: [Docker] Create image request
Create the structure to receive the `CreateImageRequest` and publish the `AvailableImages` property.

<details>

### Create image request

```json
{
  "interface_name": "io.edgehog.devicemanager.apps.CreateImageRequest",
  "version_major": 0,
  "version_minor": 1,
  "type": "datastream",
  "ownership": "server",
  "aggregation": "object",
  "mappings": [
    {
      "endpoint": "/image/id",
      "type": "string"
    },
    {
      "endpoint": "/image/repo",
      "type": "string"
    },
    {
      "endpoint": "/image/name",
      "type": "string"
    },
    {
      "endpoint": "/image/tag",
      "type": "string"
    }
  ]
}
```

### Available Images

```json
{
  "interface_name": "io.edgehog.devicemanager.apps.AvailableImages",
  "version_major": 0,
  "version_minor": 1,
  "type": "properties",
  "ownership": "device",
  "mappings": [
    {
      "allow_unset": true,
      "endpoint": "/%{image_id}/pulled",
      "type": "boolean"
    }
  ]
}
```

</details>

joshuachp:
Closes #277

### DIFF
--- a/edgehog-device-runtime-containers/src/docker/image.rs
+++ b/edgehog-device-runtime-containers/src/docker/image.rs
@@ -78,7 +78,7 @@ pub enum ConversionError {
 }
 
 /// Docker image struct.
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, Eq, Hash)]
 pub struct Image<S> {
     /// Id of the crated image.
     pub id: Option<String>,
@@ -307,6 +307,28 @@ impl<S: Display> Display for Image<S> {
         }
 
         write!(f, " {}", self.reference)
+    }
+}
+
+impl<S1, S2> PartialEq<Image<S2>> for Image<S1>
+where
+    S1: PartialEq<S2>,
+{
+    fn eq(
+        &self,
+        Image {
+            id,
+            reference,
+            registry_auth,
+        }: &Image<S2>,
+    ) -> bool {
+        let eq_registry_auth = match (&self.registry_auth, registry_auth) {
+            (Some(s1), Some(s2)) => *s1 == *s2,
+            (Some(_), None) | (None, Some(_)) => false,
+            (None, None) => true,
+        };
+
+        self.id.eq(id) && self.reference.eq(reference) && eq_registry_auth
     }
 }
 
@@ -583,5 +605,76 @@ mod tests {
         };
         let cred = image.docker_credentials().unwrap().unwrap();
         assert_eq!(cred, exp)
+    }
+
+    #[tokio::test]
+    async fn inspect_or_create_hello_world() {
+        let docker = docker_mock!(Client::connect_with_local_defaults().unwrap(), {
+            use futures::{stream, StreamExt};
+            let mut mock = Client::new();
+            let mut seq = mockall::Sequence::new();
+
+            mock.expect_remove_image()
+                .withf(|name, _, _| name == "docker.io/library/nginx:1.27.2-bookworm-perl")
+                .once()
+                .in_sequence(&mut seq)
+                .returning(|_, _, _| Err(crate::tests::not_found_response()));
+
+            mock.expect_create_image()
+                .withf(|options, _, _| {
+                    options.as_ref().map_or(false, |opt| {
+                        opt.from_image == "docker.io/library/nginx:1.27.2-bookworm-perl"
+                    })
+                })
+                .once()
+                .in_sequence(&mut seq)
+                .returning(|_, _, _| stream::empty().boxed());
+
+            mock.expect_inspect_image()
+                .withf(|name| name == "docker.io/library/nginx:1.27.2-bookworm-perl")
+                .once()
+                .in_sequence(&mut seq)
+                .returning(|_| {
+                    Ok(ImageInspect {
+                        id: Some(
+                            "sha256:d2c94e258dcb3c5ac2798d32e1249e42ef01cba4841c2234249495f87264ac5a".to_string(),
+                        ),
+                        ..Default::default()
+                    })
+                });
+
+            mock.expect_inspect_image()
+                .withf(|name| name == "sha256:d2c94e258dcb3c5ac2798d32e1249e42ef01cba4841c2234249495f87264ac5a")
+                .once()
+                .in_sequence(&mut seq)
+                .returning(|_| {
+                    Ok(ImageInspect {
+                        id: Some(
+                            "sha256:d2c94e258dcb3c5ac2798d32e1249e42ef01cba4841c2234249495f87264ac5a".to_string(),
+                        ),
+                        ..Default::default()
+                    })
+                });
+
+            mock
+        });
+
+        let mut image = Image::new("docker.io/library/nginx:1.27.2-bookworm-perl", None);
+
+        image.remove(&docker).await.unwrap();
+
+        let created = image
+            .inspect_or_create(&docker)
+            .await
+            .expect("failed to poll image");
+
+        assert!(created);
+
+        let created = image
+            .inspect_or_create(&docker)
+            .await
+            .expect("failed to poll image");
+
+        assert!(!created);
     }
 }
--- a/edgehog-device-runtime-containers/src/docker/image.rs
+++ b/edgehog-device-runtime-containers/src/docker/image.rs
@@ -581,5 +581,7 @@ mod tests {
             password: Some("passwd".to_string()),
             ..Default::default()
         };
+        let cred = image.docker_credentials().unwrap().unwrap();
+        assert_eq!(cred, exp)
     }
 }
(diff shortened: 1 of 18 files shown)


harlem88 (changes requested):

edgehog-device-runtime-docker/src/image.rs
```diff
@@ -0,0 +1,555 @@
...
+    pub async fn remove(
+        self,
+        client: &Client,
+    ) -> Result<Option<Vec<ImageDeleteResponseItem>>, ImageError>
+    where
+        S: Debug + Display,
+    {
+        let id = self.to_string();
+
+        debug!("removind image: {id}");
```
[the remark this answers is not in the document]
joshuachp: Done

edgehog-device-runtime-docker/src/service.rs
```diff
@@ -0,0 +1,386 @@
...
+impl<T> From<T> for ServiceError
+where
+    T: Into<DockerError>,
+{
+    fn from(value: T) -> Self {
+        ServiceError::Docker(value.into())
+    }
+}
+
+/// Manges the state of the Nodes
```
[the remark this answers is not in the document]
joshuachp: Done

edgehog-device-runtime-docker/src/service.rs
```diff
@@ -0,0 +1,386 @@
...
+        });
+
+        node.store(&self.device).await?;
+
+        Ok(())
+    }
+
+    /// Will start an application
+    #[instrument(skip(self))]
+    pub async fn star(&mut self, id: &str) -> Result<(), ServiceError>
```
rgallor, line 178: `starT` ?

rgallor (changes requested):

edgehog-device-runtime-docker/src/image.rs
```diff
@@ -0,0 +1,552 @@
...
+    /// Create an image with tag and repo.
+    pub fn with_repo(name: S, tag: S, repo: S) -> Self {
+        Self {
+            name,
+            tag,
+            repo: Some(repo),
+        }
+    }
+
+    /// Get a reference to the image
```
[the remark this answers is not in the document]
joshuachp: Yes it's correct.

You can read more examples on docker tags https://docs.docker.com/reference/cli/docker/image/tag/

edgehog-device-runtime-docker/src/service.rs
```diff
@@ -0,0 +1,387 @@
...
+    async fn create<S>(
+        &mut self,
+        device: &DeviceClient<S>,
+        client: &Docker,
+    ) -> Result<(), ServiceError>
+    where
+        S: PropertyStore,
+    {
+        match &self.inner {
+            NodeType::Image { image, .. } => {
```
rgallor, line 301: same here

rgallor (changes requested):

edgehog-device-runtime-docker/src/image.rs
```diff
@@ -0,0 +1,553 @@
...
+        // Check if the image was not found
+        let inspect = match res {
+            Ok(inspect) => inspect,
+            Err(BollardError::DockerResponseServerError {
+                status_code: 404,
+                message,
+            }) => {
+                debug!("image not found: {message}");
+
+                return Ok(None);
```
harlem88, line 220: Why do you return Ok here? Is it handled in a different way ?I refer to `404` error code, seems that image was not found.

edgehog-device-runtime-docker/src/image.rs
```diff
@@ -0,0 +1,553 @@
...
+
+        match res {
+            Ok(delete_res) => Ok(Some(delete_res)),
+            Err(BollardError::DockerResponseServerError {
+                status_code: 404,
+                message,
+            }) => {
+                warn!("image not found: {message}");
+
+                Ok(None)
```
harlem88, line 257: same

harlem88 (changes requested):

edgehog-device-runtime-docker/src/image.rs
joshuachp: The error 404 is returned for an image not found

edgehog-device-runtime-docker/src/image.rs
joshuachp: Same as above, the image is not found so we don't need to remove it

edgehog-device-runtime-containers/src/requests/image.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub(crate) registry_auth: String,
+}
+
+impl From<CreateImage> for Image<String> {
+    fn from(value: CreateImage) -> Self {
+        let registry_auth = if !value.registry_auth.is_empty() {
+            Some(value.registry_auth)
+        } else {
+            None
+        };
```
rgallor, line 44: yupu could simplify:
```
let registry_auth = (!value.registry_auth.is_empty()).then_some(value.registry_auth);
```

edgehog-device-runtime-containers/src/service/mod.rs
```diff
@@ -120,27 +121,93 @@ where
...
+                    service.nodes.add_node_sync(
+                        id,
+                        |id, node_idx| {
+                            Ok(Node::with_state(id, node_idx, State::Stored(image.into())))
+                        },
+                        &[],
+                    )?;
+                }
+                None => {
+                    debug!("addming missing resource");
```
rgallor, line 142: typo

edgehog-device-runtime-containers/src/service/resource.rs
```diff
@@ -18,43 +18,83 @@
...
-    pub(super) async fn create<D>(&mut self, _id: &Id, _device: &D, _client: &Docker) -> Result<()>
+    pub(super) async fn create<D>(&mut self, id: &Id, device: &D, client: &Docker) -> Result<()>
     where
-        D: Debug + Client + Sync,
+        D: Debug + Client + Sync + 'static,
     {
-        unimplemented!()
+        match self {
+            NodeType::Image(image) => {
+                image.inspect_or_create(client).await?;
```
rgallor, line 72: It seems this is the only place where `inspect_or_create` is called, but the boolean value is not used (even though I understand it could be useful to differentiate the case in which the Image already exists or not). Is it really  necessary to return Result<bool> rather than Result<()> at the moment?

rgallor (changes requested):

harlem88 (changes requested):

edgehog-device-runtime-containers/src/service/resource.rs
joshuachp: I've wanted to use it in a test, now I've added it.

edgehog-device-runtime-containers/src/docker/image.rs
```diff
@@ -0,0 +1,680 @@
...
+            id,
+            reference,
+            registry_auth,
+        }: &Image<S2>,
+    ) -> bool {
+        let eq_registry_auth = match (&self.registry_auth, registry_auth) {
+            (Some(s1), Some(s2)) => *s1 == *s2,
+            (Some(_), None) | (None, Some(_)) => false,
+            (None, None) => true,
+        };
```
lucaato, line 329:
```suggestion
        let eq_registry_auth = match (&self.registry_auth, registry_auth) {
            (Some(s1), Some(s2)) => *s1 == *s2,
            (None, None) => true,
            _ => false,
        };
```

lucaato: I'm wondering if even a simple `&self.registry_auth == registry_auth` would work

rgallor (approved):

lucaato (approved):

edgehog-device-runtime-containers/src/docker/image.rs
joshuachp: I would prefer being explisit here in the match.

The PartialEq trait is implemented as `PartialEq<Option<T>> for Option<T>` while here we want to `PartialEq<Option<U>> for Option<T>`. So the `==` wont work.

Here the issue https://github.com/rust-lang/rust/issues/20063

edgehog-device-runtime-containers/src/docker/image.rs
```diff
@@ -0,0 +1,680 @@
...
+    MissingTag,
+}
+
+/// Docker image struct.
+#[derive(Debug, Clone, Eq, Hash)]
+pub struct Image<S> {
+    /// Id of the crated image.
+    pub id: Option<String>,
+    /// Reference to the image that we need to pull.
+    pub reference: S,
```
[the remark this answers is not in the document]
joshuachp: I find it generally easier this way, so I don't have to specify a lifetime on the struct for the Cow<'a>.

This way I can easily use the `&'static str, &str, String`, as needed. In the application, this will be a String; this way we don't incur in the overhead of a Cow. On the other hand, in the tests we can construct the Image from a `&'static str`

harlem88 (approved):

PR #298, opened 2023-11-03
feat(ws): implement websocket over websocket

rgallor:
We would like to establish one or more WebSocket connections on top of
an existing WebSocket channel between a device and a bridge. For
instance, this could be useful in the scenario where a remote user, who
has access to the bridge, wishes to open a remote terminal exposed by
the device on a specific port and make it accessible through a WebSocket
connection.  The implementation of this functionality is described as
follows:
- `edgehog-device-runtime-forwarder/src/messages.rs` has been extended
  to reflect the changes in the `proto.rs` file, which contains the
  [`Protobuf`](https://protobuf.dev/) definition of the data exchanged
  between the device and the bridge. Since this extension introduces the
  WebSocket message type, now `messages.rs` contains the necessary structs
  and `impl` blocks responsible for handling such messages.
- `edgehog-device-runtime-forwarder/src/connections_manager.rs` also
  handles the sending and receiving of WebSocket messages on an already
  established WebSocket connection between the device and the bridge
  (refer to the `handle_ws()` method). To establish a WebSocket
  connection, the bridge first sends a triplet of HTTP messages, with the
  third being an HTTP `upgrade` request to the WebSocket protocol. Once
  this request is received, the device creates an internal WebSocket
  connection using [`TTYD`](https://github.com/tsl0922/ttyd), a tool used
  for sharing a remote terminal over a WebSocket interface. To simplify
  the connections' collection stored within the `ConnectionsManager`
  struct, both the HTTP and WebSocket connections are stored in the same
  collection (avoiding the problem of managing two separate collections
  for each type of connection). To this extent, the WebSocket connection
  is given the same ID as the corresponding HTTP upgrade request.
- `edgehog-device-runtime-forwarder/src/connection.rs` has been split into
  submodules, containing now two builder traits: `TransportBuilder` and 
  `Transport`. These traits (declared in `mod.rs`) define the `build()` and `next()` 
  methods, respectively, which are necessary to construct and handle each 
  connection. Specifically, two builders that implement the `TransportBuilder` 
  trait have been defined to create an HTTP and a WebSocket connection, 
  respectively implemented in the `connection/http.rs` and `connection/websocket.rs`. 
  These builders are used when spawning the Tokio task responsible for handling
  a single `Connection`. The `Connection` is now a struct that is generic
  on a type that implements the `Transport` trait. This is why the `Http`
  and `WebSocket` structs both implement this trait. The `Http` struct
  sends an HTTP request and waits for a response, while the `WebSocket`
  struct continuously sends and receives data to/from TTYD.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,11 +13,12 @@ rust-version = { workspace = true }
 [workspace]
 resolver = "2"
 members = [
-      "e2e-test", "e2e-test-forwarder",
-      "edgehog-device-runtime-docker",
-      "edgehog-device-runtime-forwarder",
-      "hardware-id-service",
-      "led-manager-service",
+  "e2e-test",
+  "e2e-test-forwarder",
+  "edgehog-device-runtime-docker",
+  "edgehog-device-runtime-forwarder",
+  "hardware-id-service",
+  "led-manager-service",
 ]
 
 [workspace.package]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,11 +13,11 @@ rust-version = { workspace = true }
 [workspace]
 resolver = "2"
 members = [
-  "e2e-test",
-  "edgehog-device-runtime-docker",
-  "edgehog-device-runtime-forwarder",
-  "hardware-id-service",
-  "led-manager-service",
+      "e2e-test", "e2e-test-forwarder",
+      "edgehog-device-runtime-docker",
+      "edgehog-device-runtime-forwarder",
+      "hardware-id-service",
+      "led-manager-service",
 ]
 
 [workspace.package]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,11 +13,11 @@ rust-version = { workspace = true }
 [workspace]
 resolver = "2"
 members = [
-    "e2e-test",
-    "edgehog-device-runtime-docker",
-    "edgehog-device-runtime-forwarder",
-    "hardware-id-service",
-    "led-manager-service",
+  "e2e-test",
+  "edgehog-device-runtime-docker",
+  "edgehog-device-runtime-forwarder",
+  "hardware-id-service",
+  "led-manager-service",
 ]
 
 [workspace.package]
@@ -97,7 +97,7 @@ reqwest = "0.12.0"
 rustc_version_runtime = "0.3.0"
 rustls = "0.22.2"
 rustls-native-certs = "0.7.0"
-rustls-pemfile = "2.1.0"
+rustls-pemfile = "2.1.1"
 serde = "1.0.195"
 serde_json = "1.0.111"
 sysinfo = "0.29.11"
--- a/e2e-test-forwarder/Cargo.toml
+++ b/e2e-test-forwarder/Cargo.toml
@@ -11,6 +11,7 @@ rust-version.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+clap = { workspace = true, features = ["derive"] }
 edgehog-forwarder = { workspace = true, features = ["_test-utils"] }
 tokio = { workspace = true }
 tracing = { workspace = true }
--- a/e2e-test-forwarder/Cargo.toml
+++ b/e2e-test-forwarder/Cargo.toml
@@ -1,4 +1,4 @@
-# Copyright 2023 SECO Mind Srl
+# Copyright 2023-2024 SECO Mind Srl
 # SPDX-License-Identifier: Apache-2.0
 
 [package]
--- a/e2e-test-forwarder/Cargo.toml
+++ b/e2e-test-forwarder/Cargo.toml
@@ -1,0 +1,16 @@
+# Copyright 2023 SECO Mind Srl
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+name = "e2e-test-forwarder"
+version = "0.1.0"
+edition.workspace = true
+homepage.workspace = true
+rust-version.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+edgehog-forwarder = { workspace = true, features = ["_test-utils"] }
+tokio = { workspace = true }
+tracing = { workspace = true }
--- a/e2e-test-forwarder/Cargo.toml
+++ b/e2e-test-forwarder/Cargo.toml
@@ -6,6 +6,7 @@ name = "e2e-test-forwarder"
 version = "0.1.0"
 edition.workspace = true
 homepage.workspace = true
+publish = false
 rust-version.workspace = true
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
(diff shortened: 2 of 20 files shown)


edgehog-device-runtime-forwarder/src/connection.rs
```diff
@@ -68,27 +111,269 @@ impl DerefMut for ConnectionHandle {
...
+        debug!("response code {}", proto_res.status());
+
+        let proto_msg = ProtoMessage::Http(ProtoHttp::new(
+            id.clone(),
+            ProtoHttpMessage::Response(proto_res),
+        ));
+
+        Ok(Some(proto_msg))
+    }
+}
```
harlem88, line 214: I think you can split this file in 3 different: 

1. connection.rs ( traits, ConnectionHandle.. )
2. http.rs ( Http struct and impl)
3. webscoket.rs ( WebSocket and Impl)
 The right thing would also be to move these file to a separate `connection` module.

harlem88 (changes requested):

edgehog-device-runtime-forwarder/src/connection.rs
```diff
@@ -21,37 +21,80 @@
...
+    /// Message sent to the wrong protocol
+    WrongProtocol,
+    /// Error when receiving message on websocket connection, `{0}`.
+    WebSocket(#[from] TungError),
+    /// Trying to poll while still connecting.
+    Connecting,
+}
+
+#[derive(Debug)]
+pub(crate) enum WriteHandle {
```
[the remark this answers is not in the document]
rgallor: Because it is used to store the write side of the channel used by the Connections Manager to send WebSocket messages to the respective Connection that will handle it.
Names such as `Connection`, `Protocol` and `Transport` have already been used to define the structs/trait representing a single connection and protobuf internal representation, respectively. `CommunicationProtocol` doesn't reflect the functionality I wanted to expose.

edgehog-device-runtime-forwarder/README.md
atsetilam, line 1: The **Sequence Diagram** paragraph might be rendered as a bullet list but overall it's good already, imho.

atsetilam (approved):

edgehog-device-runtime-forwarder/README.md
rgallor: Thanks for the suggestion. In my opinion, using a bullet list would make too strict the description. I'd prefer maintaining it this way.

harlem88 (changes requested):

edgehog-device-runtime-forwarder/src/test_utils.rs
```diff
@@ -0,0 +1,357 @@
...
+    let mut buf = Vec::with_capacity(proto_msg.encoded_len());
+    proto_msg.encode(&mut buf).unwrap();
+
+    TungMessage::Binary(buf)
+}
+
+/// Create an HTTP upgrade request and wrap it into a [`tungstenite`](tokio_tungstenite::tungstenite) message.
+pub fn create_http_upgrade_req(request_id: Vec<u8>, url: &str) -> TungMessage {
+    let url = Url::parse(url).expect("failed to pars Url");
+    let port = url.port().expect("nonexistent port").into();
```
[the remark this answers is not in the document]
rgallor: Yes I agree with you, even though it is only a test function (and it is used only once). Btw, I can change it

edgehog-device-runtime-forwarder/src/collection.rs
```diff
@@ -41,22 +43,70 @@ impl Connections {
...
+        // the HTTP message can't be an http response
+        let http_req = http_msg.into_req().ok_or_else(|| {
+            error!("Http response should not be sent by Edgehog");
+            Error::WrongMessage(request_id.clone())
+        })?;
+
+        // before executing the HTTP request, check if it is an Upgrade request.
+        // if so, handle it properly.
+        if http_req.is_ws_upgrade() {
+            info!("Connection upgrade");
```
harlem88, line 63: Change to debug level and improve the log description.
```suggestion
            debug!("Connection upgrade ...");
```

harlem88 (changes requested):

edgehog-device-runtime-forwarder/src/collection.rs
```diff
@@ -41,22 +43,70 @@ impl Connections {
...
+        // the HTTP message can't be an http response
+        let http_req = http_msg.into_req().ok_or_else(|| {
+            error!("Http response should not be sent by Edgehog");
+            Error::WrongMessage(request_id.clone())
+        })?;
+
+        // before executing the HTTP request, check if it is an Upgrade request.
+        // if so, handle it properly.
+        if http_req.is_ws_upgrade() {
+            debug!("Connection upgrade...");
```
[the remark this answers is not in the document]
rgallor: ops, sorry

Cargo.toml
```diff
@@ -13,11 +13,11 @@ rust-version = { workspace = true }
 [workspace]
 resolver = "2"
 members = [
-    "e2e-test",
-    "edgehog-device-runtime-docker",
-    "edgehog-device-runtime-forwarder",
-    "hardware-id-service",
-    "led-manager-service",
+      "e2e-test", "e2e-test-forwarder",
```
joshuachp, line 16: Format this file

e2e-test-forwarder/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+# Copyright 2023-2024 SECO Mind Srl
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+name = "e2e-test-forwarder"
+version = "0.1.0"
+edition.workspace = true
+homepage.workspace = true
+rust-version.workspace = true
```
joshuachp, line 9: Add a field for publish = false

e2e-test-forwarder/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+clap = { workspace = true, features = ["derive"] }
+edgehog-forwarder = { workspace = true, features = ["_test-utils"] }
+tokio = { workspace = true }
+tracing = { workspace = true }
```
joshuachp, line 17: Add and init the subscriber or the traces wouldn't be visible

joshuachp (changes requested):

edgehog-device-runtime-forwarder/Cargo.toml
```diff
@@ -42,8 +29,12 @@ tokio = { workspace = true }
 tokio-tungstenite = { workspace = true, features = ["rustls-tls-native-roots"] }
 tracing = { workspace = true, features = ["log"] }
 url = { workspace = true }
+uuid = { workspace = true, optional = true }
```
joshuachp, line 32: Remove this if optional and not used or only a dev dependency

edgehog-device-runtime-forwarder/src/collection.rs
```diff
@@ -41,22 +43,70 @@ impl Connections {
...
+        let ProtoHttp {
+            request_id,
+            http_msg,
+        } = http;
+
+        // the HTTP message can't be an http response
+        let http_req = http_msg.into_req().ok_or_else(|| {
+            error!("Http response should not be sent by Edgehog");
+            Error::WrongMessage(request_id.clone())
+        })?;
```
joshuachp, line 58:
```suggestion
        let Some(http_req) = http_msg.into_req() else {
            error!("Http response should not be sent by Edgehog");
            
            return Err(Error::WrongMessage(request_id));
        };
```

harlem88 (approved):

PR #325, opened 2024-01-11
refactor(path): use path instead of string

joshuachp:
Use the `Path{Buf}` struct instead of strings to simplify the code.


rgallor (approved):

harlem88 (approved):
LGTM

PR #344, opened 2024-03-07
feat(MSRV): bump Rust version to 1.72.0

joshuachp:


.github/workflows/check.yaml
```diff
@@ -112,7 +112,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        msrv: [1.66.1]
+        msrv: [1.72.1]
```
harlem88, line 115:
```suggestion
        msrv: [1.72.0]
```

harlem88 (changes requested):

.github/workflows/check.yaml
joshuachp: Fixed

PR #351, opened 2024-03-13
feat(store): use SqliteStore, not MemoryStore

rgallor:


src/data/astarte_message_hub_node.rs
```diff
@@ -236,10 +242,17 @@ mod tests {
 
         let (server_handle, drop_sender, port) = run_local_server(msg_hub).await;
 
+        let (store, tmp_store_path) = create_tmp_store().await;
+        let tmp_inrtefaces_path =
+            TempDir::new("edgehog-tmp-interfaces").expect("failed to create interface tmp dir");
+
```
joshuachp, line 248: We can use the store directory here

joshuachp: Since we know it exists and is empty (without JSON files)

src/data/astarte_message_hub_node.rs
```diff
@@ -260,10 +273,17 @@ mod tests {
 
         let (server_handle, drop_sender, port) = run_local_server(msg_hub).await;
 
+        let (store, tmp_store_path) = create_tmp_store().await;
+        let tmp_inrtefaces_path =
+            TempDir::new("edgehog-tmp-interfaces").expect("failed to create interface tmp dir");
```
joshuachp, line 278: Also here the store path

joshuachp (changes requested):
LGTM.

src/data/astarte_message_hub_node.rs
rgallor: Sure it make sense. Since it is a test, we can do this way

joshuachp:
Check the formatting, also the previous comment was for multiple places where you initialized a new temp directory. Not necessary to change it, but it would be nice while you are at it.

src/data/mod.rs
```diff
@@ -55,11 +59,66 @@ pub trait Subscriber {
...
+    Connect(#[source] SqliteError),
+    /// Path is not UTF-8, `{0}`
+    PathUtf8(PathBuf),
+}
+
+/// Options to connect a store to the Astarte device sdk or to the Message Hub nodes
+pub struct ConnectOptions<'i, 's> {
+    interface_dir: &'i Path,
+    store_dir: &'s Path,
+}
```
joshuachp, line 75:
```suggestion
pub struct ConnectOptions<'a> {
    interface_dir: &'a Path,
    store_dir: &'a Path,
}
```

joshuachp: They should live for at least the same time

joshuachp (changes requested):

src/data/astarte_message_hub_node.rs
```diff
@@ -354,10 +361,12 @@ mod tests {
 
         let (server_handle, drop_sender, port) = run_local_server(msg_hub).await;
 
+        let (store, tmp_store_path) = create_tmp_store().await;
```
joshuachp, line 364: Refactor this to remove the dir created before and use the tmp_store_path to write the interface file in

joshuachp: Same in the other three places

joshuachp (changes requested):

src/data/mod.rs
```diff
@@ -55,11 +59,66 @@ pub trait Subscriber {
...
+
+/// Connect to the store.
+pub async fn connect_store<P>(store_dir: P) -> Result<SqliteStore, StoreError>
+where
+    P: AsRef<Path>,
+{
+    let db_path = store_dir.as_ref().join("database.db");
+    let db_path_str = db_path
+        .to_str()
+        .ok_or_else(|| StoreError::PathUtf8(db_path.clone()))?;
```
joshuachp, line 104:
```suggestion
        .ok_or_else(|| StoreError::PathUtf8(db_path))?;
```

joshuachp (changes requested):

src/error.rs
```diff
@@ -63,6 +63,9 @@ pub enum DeviceManagerError {
     #[error("the connection was closed")]
     Disconnected,
 
+    #[error("Store error")]
```
joshuachp, line 66: The error should be more specific: `couldn't connect to the store`

joshuachp (changes requested):

joshuachp (approved):

src/data/mod.rs
```diff
@@ -55,11 +59,66 @@ pub trait Subscriber {
...
+    where
+        I: AsRef<Path>,
+        S: AsRef<Path>,
+    {
+        Self {
+            interface_dir: interface_dir.as_ref(),
+            store_dir: store_dir.as_ref(),
+        }
+    }
+}
```
harlem88, line 88: is it really useful? the store_directory doesn't appear used in other parts, so is it necessary to store it into ConnectionOptions? I think also you can remove this struct and pass only interface_directory directly to connect method.

harlem88 (changes requested):

src/main.rs
```diff
@@ -77,14 +78,20 @@ async fn main() -> Result<(), DeviceManagerError> {
             })?;
     }
 
+    let store = connect_store(&options.store_directory).await?;
```
harlem88, line 81: What do you think to move `connect_store` method into `connect` method and remove `store` from the parameters of the `connect` method?

src/data/mod.rs
rgallor: No, it is not necessary. It was meant to put together all the information related to the establishment of a connection

harlem88 (changes requested):

src/main.rs
rgallor: This would mean duplicating the code in both the `AstarteDeviceSdk` and `AstarteMessageHub` connect methods. Instead, in this way I create the store only in one place

harlem88: Okay, I don't really agree with you, but leave it at that for now.

PR #371, opened 2024-05-28
Forward port v0.7.2

rgallor:


joshuachp (changes requested):
We should add a commit to fix tarpaulin here.

Also, shouldn't we also release a new version of the forwarder? Since the dependencies have been updated

harlem88 (approved):

PR #373, opened 2024-06-04
chore(deps): bump forwarder proto to v0.1.0

rgallor:


harlem88:
Add bump this to the changelog, so we can notify to the user we removed the alpha deps of forwarder proto

joshuachp (approved):

harlem88 (approved):

PR #400, opened 2024-09-23
refactor(tasks): separate and handle the different runtime tasks

joshuachp:
Refactor the runtime to better handle error while gathering and sending telemetry. Group the tasks in a single `JoinSet` so we can catch errors and exit properly. 

Separate the parsing of the Astarte events from the telemetry or OTA functions.


rgallor (changes requested):

src/telemetry/mod.rs
```diff
@@ -47,400 +58,481 @@ pub(crate) mod wifi_scan;
...
-            _output = Telemetry::data_send_loop(interface_name, period, communication_channel) => {debug!("data_send_loop ended")},
-            _ = kill_switch.recv() => {debug!("Kill switch triggered")},
-        }
+#[derive(Debug)]
+pub struct Telemetry<T> {
+    client: T,
+    configs: HashMap<TelemetryInterface, TaskConfig>,
+    cancellation: CancellationToken,
+    tasks: HashMap<TelemetryInterface, CancellationToken>,
+    file_state: FileStateRepository<Vec<TelemetryInterfaceConfig<'static>>>,
```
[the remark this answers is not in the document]
joshuachp: It's only used here

src/telemetry/system_info.rs
```diff
@@ -1,73 +1,135 @@
...
+            part_number,
         }
+
+        //let mut ret: HashMap<String, AstarteType> = HashMap::new();
+        //if let Some(f) = seral_number {
+        //    ret.insert("/serialNumber".to_owned(), f.into());
+        //}
+        //if let Some(f) = part_number {
+        //    ret.insert("/partNumber".to_owned(), f.into());
+        //}
```
harlem88, line 68: Remove commented code.
```suggestion
```

harlem88 (changes requested):

src/controller/message.rs
```diff
@@ -0,0 +1,515 @@
...
+            "DoubleBlink60Seconds" => Ok(Self::Double),
+            "SlowBlink60Seconds" => Ok(Self::Slow),
+            _ => {
+                error!("unrecognize LedBehavior behavior value {value}");
+
+                Err(TypeError::Conversion)
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
joshuachp: I think separating in different modules scatters the parsing of the data event. 

Nonetheless, I have done it, let me know what you think

src/ota/mod.rs
```diff
@@ -104,3 +126,2053 @@ impl Default for DeployStatus {
...
+            | OtaStatus::Deploying(ota_request, _)
+            | OtaStatus::Deployed(ota_request)
+            | OtaStatus::Rebooting(ota_request)
+            | OtaStatus::Success(ota_request)
+            | OtaStatus::Error(_, ota_request) => Some(ota_request),
+            OtaStatus::Failure(_, ota_request) => ota_request.as_ref(),
+            _ => None,
+        }
+    }
+}
```
rgallor, line 263: this is similar to `ota_id`. If not necessary you could remove it and use the othert

src/ota/mod.rs
```diff
@@ -104,3 +126,2053 @@ impl Default for DeployStatus {
...
+    U: StateRepository<PersistentState>,
+{
+    type Msg = OtaMessage;
+
+    fn task() -> &'static str {
+        "ota"
+    }
+
+    async fn init(&mut self) -> stable_eyre::Result<()> {
+        self.ota_status = OtaStatus::Rebooted;
```
harlem88, line 292: The first status should always be `Idle`. The `Rebooted` status indicates that an OTA is in progress and the system has entered the Rebooted status. If you set the `Rebooted` status as the default, it implies that after a reboot, there is always an OTA operation pending. However, a reboot can originate from different system behaviors.

src/ota/mod.rs
```diff
@@ -104,3 +126,2053 @@ impl Default for DeployStatus {
...
+
+        // Not cancellable after a reboot, this is just a placeholder
+        let canel = CancellationToken::new();
+
+        self.handle_ota_update(canel).await;
```
harlem88, line 297:
```suggestion
        let cancel = CancellationToken::new();

        self.handle_ota_update(cancel).await;
```

src/ota/mod.rs
```diff
@@ -104,3 +126,2053 @@ impl Default for DeployStatus {
...
+        tx_publisher: mpsc::Sender<OtaStatus>,
+        flag: OtaInProgress,
+        system_update: T,
+        state_repository: U,
+    ) -> Result<Self, DeviceManagerError> {
+        Ok(Ota {
+            system_update,
+            state_repository,
+            download_file_path: opts.download_directory.clone(),
+            ota_status: OtaStatus::Rebooted,
```
harlem88, line 333: same above
```suggestion
            ota_status: OtaStatus::Idle,
```

src/ota/mod.rs
```diff
@@ -104,3 +126,2053 @@ impl Default for DeployStatus {
...
+pub struct Ota<T, U>
+where
+    T: SystemUpdate,
+    U: StateRepository<PersistentState>,
+{
+    pub system_update: T,
+    pub state_repository: U,
+    pub download_file_path: PathBuf,
+    pub ota_status: OtaStatus,
+    pub flag: OtaInProgress,
```
harlem88, line 275: I'm not sure about introducing another flag to obtain information that you can already get from status, and keep in sync with the status. Perhaps would  be better to add a function `is_in_progress` that returns a boolean from the `ota_status`. What do you think?

harlem88 (changes requested):

src/ota/mod.rs
joshuachp: If you see here in master:

https://github.com/edgehog-device-manager/edgehog-device-runtime/blob/aff8ebd19811354c1cc8d4ce1c8d494f8eafac91/src/lib.rs#L103

During the creation of the DeviceRuntime struct, we ensure that the pending ota is finished. 

This meeasn sending an EnsurePendingOta message to the ota task: 

https://github.com/edgehog-device-manager/edgehog-device-runtime/blob/aff8ebd19811354c1cc8d4ce1c8d494f8eafac91/src/ota/ota_handle.rs#L158-L164

Which starts the ota event loop with the status Rebooted.

So the initial status should actually be Rebooted since we have to ensure we didn't reboot for an OTA update. After we have checked the state repository, we return a NoPending OTA (like what master does with the ensure_no_pending_ota) and return to the Idle state.

src/ota/mod.rs
joshuachp: The flag/status need to be shared with the OtaHandle, this can be achieved by wrapping the status in a Mutex, but the only information we want to know is if the OtaStatus is not Idle.

I think this is better achieved with the AtomicBool (the OtaInProgress struct), but both are viable options.

Just tell me what you prefer

src/telemetry/hardware_info.rs
```diff
@@ -18,38 +18,139 @@
...
+    where
+        T: Publisher,
+    {
+        publish(client, INTERFACE, "/cpu/architecture", self.architecture).await;
+
+        if let Some(model) = self.model {
+            publish(client, INTERFACE, "/cpu/model", model).await;
+        } else {
+            debug!("missing cpu model");
+        }
```
rgallor, line 94: In `system_info.rs`, `os_release.rs` and other files where you perform the `send` operation you don't put any debug info. Of course this is not an error, still I would do similar things the same way (either use match or if-let-else and either decide to debug or not)

rgallor (changes requested):

src/ota/mod.rs
```diff
@@ -104,3 +126,2053 @@ impl Default for DeployStatus {
...
+            self.publish_status(self.ota_status.clone()).await;
+
+            // If the status is idle there is no OTA in progress
+            if self.ota_status == OtaStatus::Idle {
+                self.flag.set_in_progress(false);
+
+                break;
+            } else {
+                self.flag.set_in_progress(true);
+            }
```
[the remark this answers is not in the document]
joshuachp: I like it in the same expression if expression to make it clearer that we set it if the status is not idle. 

Actually, it's probably better to separate it to a different function.

src/telemetry/hardware_info.rs
joshuachp: I think it's the other way around. The system_info didn't have it, now I added it.

Comment if you find other that you think are missing.

e2e-test/Cargo.toml
```diff
@@ -26,13 +26,17 @@ rust-version = { workspace = true }
 
 [dependencies]
 astarte-device-sdk = { workspace = true, features = ["derive"] }
+clap = { workspace = true, features = ["derive", "env"] }
```
rgallor, line 29: test

joshuachp: bruh

rgallor (changes requested):

src/ota/mod.rs
```diff
@@ -104,3 +126,2108 @@ impl Default for DeployStatus {
...
+        info!("Completed signal! {:?}", signal);
+
+        match signal {
+            0 => {
+                info!("Update successful");
+
+                OtaStatus::Deployed(ota_request)
+            }
+            _ => {
+                let message = format!("Update failed with signal {signal}",);
```
rgallor, line 603: remove ending `,`

src/ota/mod.rs
```diff
@@ -104,3 +126,2108 @@ impl Default for DeployStatus {
...
+
+        let booted_slot = self.system_update.boot_slot().await.map_err(|error| {
+            let message = "Unable to identify the booted slot";
+            error!("{message}: {error}");
+            OtaError::Internal(message)
+        })?;
+
+        if state.slot == booted_slot {
+            let message = "Unable to switch slot";
+            return Err(OtaError::SystemRollback(message));
```
rgallor, line 683: Can't you just inline the message here since you are not logging?

src/ota/mod.rs
```diff
@@ -104,3 +126,2108 @@ impl Default for DeployStatus {
...
+            OtaStatus::Deployed(ota_request) => self.reboot(ota_request).await,
+            OtaStatus::Rebooted => self.check_reboot().await,
+            OtaStatus::Error(ota_error, ota_request) => {
+                OtaStatus::Failure(ota_error, Some(ota_request))
+            }
+            rebooting @ OtaStatus::Rebooting(_) => rebooting,
+            OtaStatus::NoPendingOta | OtaStatus::Success(_) | OtaStatus::Failure(_, _) => {
+                OtaStatus::Idle
+            }
+            OtaStatus::Idle => OtaStatus::Idle,
```
rgallor, line 725: Is this put separately from the previous to emphasize that, if the state is Idle, it will remain Idle once `next` is called? Otherwise you could merge it with the previous ones

rgallor (changes requested):

src/telemetry/event.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn try_from(value: AstarteType) -> Result<Self, Self::Error> {
+        let secs = i64::try_from(value).map(|i| match u64::try_from(i) {
+            Ok(secs) => secs,
+            Err(_) => {
+                warn!("Telemetry period seconds value too big {i}, capping to u64::MAX");
+
+                u64::MAX
+            }
+        })?;
+
```
rgallor, line 87: you could simplify with the following: 
```rust 
let secs = i64::try_from(value).map(|i| {
    u64::try_from(i).unwrap_or_else(|_| {
        warn!("Telemetry period seconds value too big {i}, capping to u64::MAX");
        u64::MAX
    })
})?;
```

rgallor (changes requested):

src/ota/mod.rs
joshuachp: Added log

src/ota/mod.rs
joshuachp: Yes, it was to make it clearer, but it's the same anyway. Joined with the previous pattern

src/lib.rs
```diff
@@ -19,22 +19,17 @@
...
 
-use crate::data::{Publisher, Subscriber};
 use crate::error::DeviceManagerError;
-use crate::ota::ota_handler::OtaHandler;
-use crate::telemetry::{TelemetryMessage, TelemetryPayload};
+
+pub use self::controller::Runtime;
 
 mod commands;
+pub mod controller;
```
harlem88, line 32: This can be keep private.
```suggestion
mod controller;
```

src/ota/mod.rs
harlem88: Start from Idle status also here.

src/ota/ota_handler.rs
```diff
@@ -54,368 +51,193 @@ pub struct OtaEvent {
     pub message: String,
 }
 
-struct OtaStatusMessage {
-    status_code: String,
-    message: String,
+pub(crate) struct OtaStatusMessage {
+    pub(crate) status_code: String,
+    pub(crate) message: String,
```
harlem88, line 56: Since the `ota_handler` is already private to the crate, you can remove the `pub(crate)` from struct and method declaration. 
```suggestion
pub struct OtaStatusMessage {
    pub status_code: String,
    pub message: String,
```

harlem88 (changes requested):

src/telemetry/hardware_info.rs
```diff
@@ -18,38 +18,139 @@
...
+                    "couldn't get the memory info: {}",
+                    stable_eyre::Report::new(err)
+                );
 
-    Ok(ret)
+                return mem;
+            }
+        };
+
+        if let Ok(mem_total) = i64::try_from(mem_info.mem_total) {
```
rgallor, line 132: what do you think of the following refactor?
```rust
mem.total_bytes = i64::try_from(mem_info.mem_total).ok().or_else(|| {
            error!(
                "mem total too big to be sent to astarte: {}",
                mem_info.mem_total
            );
            mem.total_bytes
        });
```

rgallor (changes requested):

rgallor (approved):
Apart for the unresolved comments, it LGTM 🔥

src/ota/mod.rs
```diff
@@ -104,3 +126,2109 @@ impl Default for DeployStatus {
...
+        tx_publisher: mpsc::Sender<OtaStatus>,
+        flag: OtaInProgress,
+        system_update: T,
+        state_repository: U,
+    ) -> Result<Self, DeviceManagerError> {
+        Ok(Ota {
+            system_update,
+            state_repository,
+            download_file_path: opts.download_directory.clone(),
+            ota_status: OtaStatus::Rebooted,
```
harlem88, line 383: Why is this still in Rebooted, and not in Idle?

joshuachp: My mistake, now it's fixed

harlem88 (approved):

PR #420, opened 2024-10-24
refactor(container): improve the Resource Id

joshuachp:
- Tag the container resource Id with the resource type
- Make the string part of the Id a UUID

### DIFF
--- a/edgehog-device-runtime-containers/src/service/mod.rs
+++ b/edgehog-device-runtime-containers/src/service/mod.rs
@@ -19,7 +19,6 @@
 //! Service to receive and handle the Astarte events.
 
 use std::{
-    borrow::Borrow,
     fmt::{Debug, Display},
     ops::Deref,
     sync::Arc,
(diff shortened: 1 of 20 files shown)


edgehog-device-runtime-containers/src/service/mod.rs
```diff
@@ -533,7 +485,35 @@ impl AsRef<str> for Id {
...
+    Image = 0,
+    /// Volume resource.
+    Volume = 1,
+    /// Network resource.
+    Network = 2,
+    /// Container resource.
+    Container = 3,
+    /// Deployment resource.
+    Deployment = 4,
+}
```
rgallor: Why are you making explicit the numbers associated to each enum value?

rgallor (changes requested):

edgehog-device-runtime-containers/src/service/mod.rs
joshuachp: To be stable and serializable as numbers

rgallor (approved):

harlem88 (approved):

PR #422, opened 2024-10-28
Moved ota, udev and wifiscanner to optional features

### LINKED ISSUE #414: Enable some modules by Rust feature
Some device-runtime features are not supported by all platforms where can install it, so we can move these part of code or module below a Rust feature.

Currently we can move below features:
- udev
- wif_scanner
- rauc

https://github.com/edgehog-device-manager/edgehog-device-runtime/blob/5945f7714d3e99db4bbbab000e7731c8756a466f/src/telemetry/mod.rs#L48

https://github.com/edgehog-device-manager/edgehog-device-runtime/blob/5945f7714d3e99db4bbbab000e7731c8756a466f/src/telemetry/mod.rs#L56

https://github.com/edgehog-device-manager/edgehog-device-runtime/blob/5945f7714d3e99db4bbbab000e7731c8756a466f/src/ota/ota_handler.rs#L101

``` toml
[features]
udev = ["dep:udev"]
wifiscanner = ["dep:wifiscanner"]
rauc = ["dep:zbus"]
```
``` rust
#[cfg(feature = "udev")]
pub(crate) mod net_interfaces;
...
#[cfg(feature = "wifiscanner")]
pub(crate) mod wifi_scan;
```
``` rust
 #[cfg(feature = "rauc")]
 let system_update = OTARauc::connect().await?;
 ...
```

lucaato:
Fixes #414

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ rust-version = "1.72.0"
 astarte-device-sdk = { workspace = true, features = ["derive"] }
 async-trait = { workspace = true }
 bytes = { workspace = true }
+cfg-if = { workspace = true }
 clap = { workspace = true, features = ["derive", "env"] }
 displaydoc = { workspace = true }
 edgehog-forwarder = { workspace = true, optional = true }
@@ -54,11 +55,11 @@ tokio-util = { workspace = true }
 toml = { workspace = true }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
-udev = { workspace = true }
+udev = { workspace = true, optional = true }
 url = { workspace = true }
 uuid = { workspace = true, features = ["v5", "v4", "serde"] }
-wifiscanner = { workspace = true }
-zbus = { workspace = true, default-features = false, features = ["tokio"] }
+wifiscanner = { workspace = true, optional = true }
+zbus = { workspace = true, optional = true, default-features = false, features = ["tokio"] }
 
 [build-dependencies]
 rustc_version = { workspace = true }
@@ -72,9 +73,13 @@ tokio = { workspace = true, features = ["test-util"] }
 tokio-stream = { workspace = true, features = ["net"] }
 
 [features]
+default = ["udev", "wifiscanner", "zbus"]
 message-hub = ["astarte-device-sdk/message-hub"]
 systemd = ["dep:systemd"]
 forwarder = ["dep:edgehog-forwarder"]
+udev = ["dep:udev"]
+wifiscanner = ["dep:wifiscanner"]
+zbus = ["dep:zbus"]
 
 [workspace.dependencies]
 astarte-device-sdk = "0.9.1"
--- a/src/data/astarte_device_sdk_lib.rs
+++ b/src/data/astarte_device_sdk_lib.rs
@@ -30,7 +30,6 @@ use serde::Deserialize;
 use tokio::task::JoinSet;
 use url::Url;
 
-use crate::device::DeviceProxy;
 use crate::repository::file_state_repository::{FileStateError, FileStateRepository};
 use crate::repository::StateRepository;
 
@@ -40,6 +39,7 @@ pub enum DeviceSdkError {
     /// missing device ID
     MissingDeviceId,
     /// couldn't get the hardware id from DBus
+    #[cfg(all(feature = "zbus", target_os = "linux"))]
     Zbus(#[from] zbus::Error),
     /// couldn't pair device to Astarte
     Pairing(#[from] PairingError),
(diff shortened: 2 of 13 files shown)


Cargo.toml
```diff
@@ -75,6 +76,14 @@ tokio-stream = { workspace = true, features = ["net"] }
...
 systemd = ["dep:systemd"]
 forwarder = ["dep:edgehog-forwarder"]
+udev = ["dep:udev"]
+wifiscanner = ["dep:wifiscanner"]
+rauc = ["dep:zbus"]
+upower = ["dep:zbus"]
+cellular = ["dep:zbus"]
+battery = ["dep:zbus"]
+led = ["dep:zbus"]
+dbus_hid = ["dep:zbus"]
```
harlem88, line 86: mmm, I don't know if we want to specify all as features, perhaps we can use directly zbus as discriminant.
For example enable led_behavior only checking zbus.
``` rust
#[cfg(feature = "zbus")]
Led(led_behavior::LedEvent),
```
What do you think?

harlem88 (changes requested):

Cargo.toml
lucaato: Yes, it makes sense, in your opinion which features should be enabled by default by zbus ?
Should I merge them all ? I would keep the `dbus_hid` separated since if disabled it requires the user to set a device id

harlem88: Since we have to keep retro-compatibility all these zbus features should be enabled by default.
I think also `dbus_hid` can be merged, an user can pass a device_ID via config as other configurations. 
I think we can start simple keep all below `dep:zbus`.

joshuachp: We could also add a cfg!(not([target_os](https://doc.rust-lang.org/reference/conditional-compilation.html#target_os) = "linux")) to keep it as default but permit the build on other platforms

src/ota/ota_handler.rs
```diff
@@ -36,11 +36,12 @@ use crate::ota::rauc::OTARauc;
 use crate::ota::OtaError;
 use crate::ota::{Ota, OtaId, OtaStatus};
 use crate::repository::file_state_repository::FileStateRepository;
-use crate::MAX_OTA_OPERATION;
 
 use super::event::{OtaOperation, OtaRequest};
 use super::PersistentState;
 
+const MAX_OTA_OPERATION: usize = 2;
```
harlem88, line 43: Why do you redefine `MAX_OTA_OPERATION`?

harlem88 (changes requested):

src/ota/ota_handler.rs
lucaato: I moved it from the main lib module to the OTA module since it makes more sense to have it in the same place where we're using it. Given that it could be disabled, it avoids unused warnings when the ̀zbus̀ feature is disabled.

Cargo.toml
```diff
@@ -75,6 +76,9 @@ tokio-stream = { workspace = true, features = ["net"] }
 message-hub = ["astarte-device-sdk/message-hub"]
 systemd = ["dep:systemd"]
 forwarder = ["dep:edgehog-forwarder"]
+udev = ["dep:udev"]
```
harlem88, line 79: What do you think to keep these features as default? Also to keep retro-compatibility with who already include our device-runtime in their image build system.
default = ["udev", "wifiscanner", "zbus"]

harlem88 (changes requested):

harlem88 (approved):

src/data/astarte_device_sdk_lib.rs
```diff
@@ -72,9 +72,16 @@ impl AstarteDeviceSdkConfigOptions {
...
-        hardware_id_from_dbus()
-            .await?
-            .ok_or(DeviceSdkError::MissingDeviceId)
+        cfg_if::cfg_if! {
+            if #[cfg(all(feature = "zbus", target_os = "linux"))] {
+                hardware_id_from_dbus()
+                    .await?
+                    .ok_or(DeviceSdkError::MissingDeviceId)
+            }
+            else {
```
joshuachp, line 81:
```suggestion
            } else {
```

src/telemetry/sender.rs
```diff
@@ -94,7 +94,14 @@ impl<T> Task<T> {
                 StorageUsage::read().send(&self.client).await;
             }
             TelemetryInterface::BatteryStatus => {
-                battery_status::send_battery_status(&self.client).await;
+                cfg_if::cfg_if! {
+                    if #[cfg(all(feature = "zbus", target_os = "linux"))] {
+                        super::battery_status::send_battery_status(&self.client).await;
+                    }
+                    else {
```
joshuachp, line 101:
```suggestion
                    } else {
```

joshuachp (changes requested):
nit: some macro formatting was weird

joshuachp (approved):

PR #428, opened 2024-10-29
Use ring as default crypto provider

harlem88:
Since we can have different versions of rustls, we decide to install ring as default provider to avoid the `no crypto provider install` issue generated by two different version of rustls used by the application.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ edgehog-forwarder = { workspace = true, optional = true }
 futures = { workspace = true }
 procfs = { workspace = true }
 reqwest = { workspace = true, features = ["stream"] }
+rustls = { workspace = true, features = ["ring","logging","std","tls12"] }
 serde = { workspace = true }
 serde_json = { workspace = true }
 stable-eyre = { workspace = true }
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -54,6 +54,11 @@ async fn main() -> stable_eyre::Result<()> {
         )
         .try_init()?;
 
+    // Use ring as default crypto provider
+    rustls::crypto::ring::default_provider()
+        .install_default()
+        .expect("Failed to install rustls crypto provider");
```
joshuachp, line 60:
```suggestion
        .wrap_err("Failed to install rustls crypto provider")?;
```

So we have the full error stack in eyre

joshuachp (changes requested):

Cargo.toml
```diff
@@ -42,6 +42,7 @@ edgehog-forwarder = { workspace = true, optional = true }
 futures = { workspace = true }
 procfs = { workspace = true }
 reqwest = { workspace = true, features = ["stream"] }
+rustls = { workspace = true, features = ["ring","logging","tls12"], default-features = false }
```
joshuachp, line 45:
```suggestion
rustls = { workspace = true, features = ["ring","logging","std","tls12"], default-features = false }
```

joshuachp: It's seems there were more default features.

joshuachp (changes requested):

joshuachp (approved):

PR #429, opened 2024-10-30
feat(containers): integrate the with the runtime

joshuachp:
Create a task to handle the container requests and commands from
Astarte.

### DIFF
--- a/src/controller/mod.rs
+++ b/src/controller/mod.rs
@@ -17,7 +17,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use actor::Actor;
-use astarte_device_sdk::{client::RecvError, FromEvent};
+use astarte_device_sdk::{client::RecvError, Client, FromEvent};
 use stable_eyre::eyre::Error;
 use tokio::{sync::mpsc, task::JoinSet};
 use tracing::{error, info};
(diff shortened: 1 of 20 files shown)


src/controller/mod.rs
```diff
@@ -104,6 +111,24 @@ impl<T> Runtime<T> {
...
+    #[cfg(feature = "containers")]
+    async fn setup_containers(
+        store_dir: &std::path::Path,
+        client: T,
+        tasks: &mut JoinSet<stable_eyre::Result<()>>,
+    ) -> Result<mpsc::Sender<Box<edgehog_containers::requests::ContainerRequest>>, DeviceManagerError>
+    where
+        T: Client + Send + Sync + 'static,
+    {
+        let (container_tx, container_rx) = mpsc::channel(8);
```
rgallor, line 123: Why don't we add here a const value, like CONTAINER_CHANNEL_SIZE?

rgallor (changes requested):

src/controller/mod.rs
joshuachp: I could, but there are many other channels initialized in this file like this, without constants.

I think the channel configurations should be refactored in another PR, maybe an issue should be opened about it if we want to make them configurable. Anyway, I believe it's separate from the current changes.

harlem88 (approved):

rgallor (approved):

PR #434, opened 2024-11-05
Fix task error report

### LINKED ISSUE #430: Check if ota_state file is valid
https://github.com/edgehog-device-manager/edgehog-device-runtime/blob/17ee1c93d3973df739f0d858dd0c566c15b9a046/src/ota/mod.rs#L644-L658

``` sh
Error: handle for ota-publisher task failed

Caused by:
    InternalError: Unable to publish ota_event: request_uuid is empty
```

harlem88:
close #430

### DIFF
--- a/src/ota/mod.rs
+++ b/src/ota/mod.rs
@@ -133,6 +133,7 @@ impl Default for DeployStatus {
 }
 
 const DOWNLOAD_PERC_ROUNDING_STEP: f64 = 10.0;
+const DEPLOY_PERC_ROUNDING_STEP: i32 = 10;
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct PersistentState {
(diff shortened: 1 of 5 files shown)


joshuachp (changes requested):

src/ota/mod.rs
```diff
@@ -353,13 +359,16 @@ where
         if self.ota_status != OtaStatus::Idle {
             error!("ota request already in progress");
 
-            return Err(OtaError::UpdateAlreadyInProgress.into());
```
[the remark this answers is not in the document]
harlem88: Yep, i made mistake with the ota_handler and I changed the return with your suggestion of `IncosistentState`.

src/ota/mod.rs
```diff
@@ -121,6 +121,9 @@ pub enum OtaError {
     /// OTA update aborted by Edgehog half way during the procedure
     #[error("Canceled")]
     Canceled,
+    #[error("InconsistentState")]
+    /// Attempted to start OTA operation while ota status is different from Idle*/
```
joshuachp, line 125:
```suggestion
    /// Attempted to start OTA operation while ota status is different from Idle
    #[error("inconsistent ota state")
```

src/ota/mod.rs
```diff
@@ -355,13 +363,12 @@ where
         if self.ota_status != OtaStatus::Idle {
             error!("ota request already in progress");
 
-            return Err(OtaError::UpdateAlreadyInProgress.into());
+            return Err(OtaError::InconsistentState.into());
+        } else {
+            self.ota_status = OtaStatus::Init(msg.ota_id);
+            self.handle_ota_update(msg.cancel).await;
         }
```
joshuachp, line 370:
```suggestion
        self.ota_status = OtaStatus::Init(msg.ota_id);
        self.handle_ota_update(msg.cancel).await;
```

src/ota/ota_handler.rs
```diff
@@ -278,6 +279,9 @@ impl From<&OtaError> for OtaStatusMessage {
                 ota_status_message.message = message.to_string()
             }
             OtaError::Canceled => ota_status_message.status_code = "Canceled".to_string(),
+            OtaError::InconsistentState => {
+                ota_status_message.status_code = "InconsistentState".to_string()
```
joshuachp, line 283: Are these also defined on the backend? Here I would return an internal error if that's the case

joshuachp: https://docs.edgehog.io/0.9/astarte_interfaces.html#event-statuscode

Here the codes are documented.

```suggestion
                ota_status_message.status_code = "Internal".to_string();
                ota_status_message.message = ota_error.to_string(); 
```

joshuachp (changes requested):

rgallor (approved):

src/ota/ota_handler.rs
harlem88: I agree with you, better to use something that it is already defined also on backed side.

joshuachp (approved):

PR #435, opened 2024-11-06
refactor(container): improve the services code

joshuachp:
- Do not error when sending data to Astarte
- Don't pass closures to the graph but the full node
- Move all the logic to the node to reduce all the function calls


edgehog-device-runtime-containers/src/service/node.rs
```diff
@@ -21,115 +21,274 @@
...
+
+        match &resource.state {
+            State::Received => {
+                self.create(device, client).await?;
+                self.start(device, client).await?;
+            }
+            State::Stored => {
+                self.create(device, client).await?;
+                self.start(device, client).await?;
+            }
```
rgallor, line 270: You can merge the 2 using `State::Received | State::Stored => {...}`

rgallor (changes requested):

harlem88 (approved):

rgallor (approved):

PR #444, opened 2024-11-15
fix(container): correctly check node existence

joshuachp:
Add the resource to previously added missing nodes.

### DIFF
--- a/edgehog-device-runtime-containers/src/service/collection.rs
+++ b/edgehog-device-runtime-containers/src/service/collection.rs
@@ -26,7 +26,7 @@ use petgraph::{
     visit::Walker,
     Direction,
 };
-use tracing::debug;
+use tracing::{debug, instrument};
 
 use super::{node::Node, resource::NodeResource, Id, ServiceError};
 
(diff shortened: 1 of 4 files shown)


harlem88 (approved):

edgehog-device-runtime-containers/src/service/collection.rs
```diff
@@ -87,6 +96,7 @@ impl NodeGraph {
         //       a mutable reference to the node while we mutate the map
         // check if missing to add relations
         let idx = match self.node(&id) {
+            // Node exists and resource is not None
             Some(node) => node.idx,
             None => {
                 // wee need to add the deps if we don't have the node or the resource is missing
```
lucaato, line 102:
```suggestion
                // we need to add the deps if we don't have the node or the resource is missing
```

rgallor (approved):

edgehog-device-runtime-containers/src/service/collection.rs
joshuachp: oh noo

lucaato: to happy to be programming in rust

lucaato (approved):

PR #451, opened 2024-11-19
refactor(state): store the resource state

joshuachp:
This removes some warnings on restart, when the resources are loaded with state stored.
Now they are loaded with the actual state.

### DIFF
--- a/edgehog-device-runtime-containers/src/service/resource.rs
+++ b/edgehog-device-runtime-containers/src/service/resource.rs
@@ -38,22 +38,19 @@ impl NodeResource {
     pub(crate) fn with_default(resource: NodeType) -> Self {
         Self::new(State::default(), resource)
     }
-
-    pub(crate) fn stored(resource: NodeType) -> Self {
-        Self {
-            state: State::Stored,
-            value: resource,
-        }
-    }
 }
 
 /// State of the object for the request.
 #[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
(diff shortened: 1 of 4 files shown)


edgehog-device-runtime-containers/src/service/resource.rs
```diff
@@ -38,30 +38,27 @@ impl NodeResource {
...
+    // The resource was is running
     Up,
 }
 
 impl Display for State {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let state = match self {
             State::Received => "Missing",
-            State::Stored => "Stored",
+            State::Published => "Stored",
```
harlem88, line 61: Do you want to keep `Stored` as Display?
```suggestion
            State::Published => "Published",
```

harlem88 (changes requested):

edgehog-device-runtime-containers/src/service/resource.rs
```diff
@@ -38,30 +38,27 @@ impl NodeResource {
...
 pub(crate) enum State {
+    // The resource was received but was not published
     #[default]
     Received,
-    Stored,
+    // The resource was published and saved to the state file
+    Published,
+    // The resource was created
     Created,
+    // The resource was is running
```
rgallor, line 53: remove `was`

rgallor (changes requested):

PR #455, opened 2024-11-22
feat(container): fetch status on restart

joshuachp:
Fetch the status on restart before starting the containers.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,6 +100,7 @@ displaydoc = "0.2.5"
 edgehog-containers = { package = "edgehog-device-runtime-containers", path = "./edgehog-device-runtime-containers", version = "=0.1.0" }
 edgehog-device-forwarder-proto = "0.1.0"
 edgehog-forwarder = { package = "edgehog-device-runtime-forwarder", path = "./edgehog-device-runtime-forwarder", version = "=0.8.1" }
+eyre = "0.6.12"
 futures = "0.3.31"
 hex = "0.4.3"
 http = "1.1.0"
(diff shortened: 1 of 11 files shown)


edgehog-device-runtime-containers/src/service/node.rs
```diff
@@ -56,6 +56,74 @@ impl Node {
...
+
+        let Some(resource) = &mut self.resource else {
+            debug!("skip missing");
+
+            return Ok(());
+        };
+
+        match resource.state {
+            State::Received | State::Published => {
+                debug!("skip resource not created yet ");
```
harlem88, line 75:
```suggestion
                debug!("skip resource not created yet");
```

harlem88 (changes requested):

Cargo.toml
```diff
@@ -100,6 +100,7 @@ displaydoc = "0.2.5"
 edgehog-containers = { package = "edgehog-device-runtime-containers", path = "./edgehog-device-runtime-containers", version = "=0.1.0" }
 edgehog-device-forwarder-proto = "0.1.0"
 edgehog-forwarder = { package = "edgehog-device-runtime-forwarder", path = "./edgehog-device-runtime-forwarder", version = "=0.8.1" }
+eyre = "0.6.12"
```
harlem88, line 103: Don't we already have `stable_eyre` & `color_eyre`?

joshuachp: Yes, but those are used in the binary and in the e2e-test. While eyre can be used in the library, since it doesn't provide a panic hook. It provides the report type to print the full error report.

harlem88 (approved):

PR #481, opened 2025-01-14
feat(store): impl the sqlite store with diesel 

joshuachp:
Create the structures to store the container states in a sql database.
Handle the migrations and the read and write handles. Make the writer
shared between threads with a Mutex and the reader cloned and owned by
the instance.


edgehog-device-runtime-store/migrations/2024-11-29-163409_initial/up.sql
```diff
@@ -0,0 +1,166 @@
...
+    status INTEGER NOT NULL,
+    driver TEXT NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS volume_driver_opts (
+    volume_id BLOB NOT NULL,
+    name TEXT NOT NULL,
+    value TEXT,
+    PRIMARY KEY (volume_id, name),
+    FOREIGN KEY (volume_id) REFERENCES volume (
```
harlem88, line 57: The table name should be `volumes`
```suggestion
    FOREIGN KEY (volume_id) REFERENCES volumes (
```

edgehog-device-runtime-store/migrations/2024-11-29-163409_initial/up.sql
```diff
@@ -0,0 +1,166 @@
...
+        id
+    ) ON DELETE CASCADE ON UPDATE CASCADE,
+    network_id BLOB NOT NULL REFERENCES networks (
+        id
+    ) ON DELETE CASCADE ON UPDATE CASCADE,
+    PRIMARY KEY (container_id, network_id)
+);
+
+CREATE TABLE IF NOT EXISTS container_missing_networks (
+    container_id BLOB NOT NULL REFERENCES container (
```
harlem88, line 93:
```suggestion
    container_id BLOB NOT NULL REFERENCES containers (
```

edgehog-device-runtime-store/migrations/2024-11-29-163409_initial/up.sql
```diff
@@ -0,0 +1,166 @@
...
+    ) ON UPDATE CASCADE ON DELETE CASCADE,
+    network_id BLOB NOT NULL,
+    PRIMARY KEY (container_id, network_id)
+);
+
+CREATE TABLE IF NOT EXISTS container_volumes (
+    container_id BLOB NOT NULL REFERENCES containers (
+        id
+    ) ON DELETE CASCADE ON UPDATE CASCADE,
+    volume_id BLOB NOT NULL REFERENCES volume (
```
harlem88, line 104:
```suggestion
    volume_id BLOB NOT NULL REFERENCES volumes (
```

edgehog-device-runtime-store/migrations/2024-11-29-163409_initial/up.sql
```diff
@@ -0,0 +1,166 @@
...
+        id
+    ) ON DELETE CASCADE ON UPDATE CASCADE,
+    PRIMARY KEY (container_id, volume_id)
+);
+
+CREATE TABLE IF NOT EXISTS container_missing_volumes (
+    container_id BLOB NOT NULL REFERENCES container (
```
harlem88, line 111:
```suggestion
    container_id BLOB NOT NULL REFERENCES containers (
```

harlem88 (changes requested):

edgehog-device-runtime-store/migrations/2024-11-29-163409_initial/up.sql
joshuachp: Thanks good catch. Unfortunately, SQLite doesn't error when there's a missing table constraint. So there is no easy to check this type of error automatically

edgehog-device-runtime-store/src/models/containers/container.rs
```diff
@@ -0,0 +1,488 @@
...
+
+    /// Returns the filtered container table by id.
+    pub fn find_id(id: &SqlUuid) -> ContainerFilterById<'_> {
+        containers::table.filter(Self::by_id(id))
+    }
+
+    /// Returns the container exists query.
+    pub fn exists(id: &SqlUuid) -> ContainerExists<'_> {
+        select(exists(Self::find_id(id)))
+    }
```
harlem88, line 85: Since each table of db is represented by a struct with these methods, What do you think to create a trait with these methods, and for the result of each methods a specific associated types?
Such as:
```rust
trait DbQuery {
 type A;
 type B;
 type C;

  pub fn by_id(id: &SqlUuid) -> A;
  pub fn find_id(id: &SqlUuid) -> B ;
  pub fn exists(id: &SqlUuid) -> C;
	
```
It seems something that you can generalize...

harlem88 (changes requested):

edgehog-device-runtime-store/src/models/containers/container.rs
joshuachp: I think it's possible. I will try it 👍🏻

joshuachp: Done, except for the `exists` I created a default implementation. I cannot come up with the correct trait bound for it.

joshuachp:
Moved to a feature branch

edgehog-device-runtime-containers/src/store/db/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+//! Persistent stores of the request issued by Astarte and resources created.
+
+#![allow(dead_code)]
+
+use std::ops::Not;
```
harlem88, line 23: Is it still used?
```suggestion
```

harlem88 (changes requested):

edgehog-device-runtime-containers/src/store/db/mod.rs
joshuachp: Yes, it's used for this `.not()` call: https://github.com/edgehog-device-manager/edgehog-device-runtime/blob/7459d3268b5cc3335ea0f8415d7327d5285f111a/edgehog-device-runtime-containers/src/store/db/mod.rs#L57

edgehog-device-runtime-containers/src/docker/container.rs
```diff
@@ -63,52 +66,52 @@ pub enum ContainerError {
...
     /// Id of the docker container.
     ///
     /// The id of the image is optional since it will be available only when the image is created.
-    pub id: Option<String>,
+    pub(crate) id: Option<String>,
     /// Assign the specified name to the container.
     ///
     /// Must match /?[a-zA-Z0-9][a-zA-Z0-9_.-]+.
-    pub name: S,
+    pub(crate) name: S,
```
harlem88, line 80: Is it not enough to put `(crate) rule` on the struct to make visible only with crate? All the fields should be not usable since the struct is not visible outer from the crate.
``` rust
pub(crate) struct Container<S>
where
    S: Hash + Eq,
{
...    
pub id: Option<String>,
...
}
```

joshuachp: Yes, that clearly works, but I think being a little bit more explicit and removing the indirection (struct private but fields public) makes it simpler to read.

I think I should have done it for all the structs/crates to begin with, but now there are a lot of them and probably a gradual change would be better.

rgallor (changes requested):

harlem88 (approved):

edgehog-device-runtime-store/src/db.rs
```diff
@@ -0,0 +1,242 @@
...
+
+        tokio::task::spawn_blocking(|| {
+            sql_query(INIT_READER)
+                .execute(&mut conn)
+                .map_err(HandleError::Query)?;
+
+            Ok(conn)
+        })
+        .await?
+    }
```
rgallor, line 165: since `establish_writer` and `establish_reader` appears to be the same apart from `INIT_WRITER` and `INIT_READER` pragma options, why don't you define a single method `establish_connection` and pass in input the pragma you want to use?

edgehog-device-runtime-store/src/db.rs
```diff
@@ -0,0 +1,242 @@
...
+        })
+    }
+
+    /// Passes the reader to a callback to execute a query.
+    pub async fn for_read<F, O>(&mut self, f: F) -> Result<O>
+    where
+        F: FnOnce(&mut SqliteConnection) -> Result<O> + Send + 'static,
+        O: Send + 'static,
+    {
+        // Take
```
rgallor, line 184:
```suggestion
        // Take the reader out of the `Handle`
```

rgallor (changes requested):

edgehog-device-runtime-store/src/models/containers/volume.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// Received from Edgehog.
+    #[default]
+    Received = 0,
+    /// The volume was acknowledged
+    Published = 1,
+    /// Created on the runtime.
+    Created = 2,
+}
+
+impl Display for VolumeStatus {
```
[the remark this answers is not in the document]
joshuachp: It's an interesting utility. 

Unfortunately, I am not fully convinced by the benefits of derive_more over a manual implementation.

The manual implementation, even though it is more boiler-platy, has the advantages of being our code, and it's easier to understand or change.

Having to add a new dependency with a proc-macro introduces more complexity and compile times to the project, while the features will be used only for a few structs.

It's a really handy utility crate, but I don't think having a macro for the Display traits or simple conversions on a few structs would overweight the cons of an additional dependency. Maybe in another PR we can reconsider it, what do you think?

lucaato (approved):

rgallor (approved):

harlem88 (approved):

PR #483, opened 2025-01-21
doc: Add NotifyAccess option to system section

harlem88:
This specifies access to the socket that should be used to listen for notifications when the “notify” service type is selected.

### DIFF
--- a/doc/os_requirements.md
+++ b/doc/os_requirements.md
@@ -59,6 +59,16 @@ feature enabled.
 cargo build --features systemd
 ```
 
+To allow systemd to receive status changes from service, you need to set the
+[NotifyAccess](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#NotifyAccess=) option to
+either `exec` or `main`.
+
+```
+[Service]
+...
+NotifyAccess=exec
+```
+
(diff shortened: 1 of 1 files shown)


doc/os_requirements.md
```diff
@@ -59,6 +59,15 @@ feature enabled.
 cargo build --features systemd
 ```
 
+To allow systemd to receive status changes from service, you need to add `NotifyAccess=exec` option
```
joshuachp, line 62:
```suggestion
To allow systemd to receive status changes from service, you need to set the `NotifyAccess` option to either `exec` or `main`. 
```

joshuachp (changes requested):

doc/os_requirements.md
joshuachp: I believe `main` should be sufficient. Since we are the only process that should be started by the service.

harlem88: We can leave both options in the doc, so the users can make a choice according to their needs.

joshuachp (approved):

joshuachp (approved):

joshuachp (approved):

PR #508, opened 2025-03-14
feat(reboot): add custom reboot

joshuachp:
Remove the direct reboot command and add a signal to the RAUC interface to wait and signal reboot errors to the device runtime.

### DIFF
--- a/src/ota/rauc.rs
+++ b/src/ota/rauc.rs
@@ -69,7 +69,7 @@ pub struct BundleInfo {
 trait Rauc {
     /// Triggers the installation of a bundle. This method call is non-blocking.
     /// After completion, the “Completed” signal will be emitted.
-    fn install_bundle(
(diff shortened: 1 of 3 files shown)


src/ota/rauc.rs
```diff
@@ -197,6 +204,31 @@ impl SystemUpdate for OTARauc<'static> {
             .await
             .map_err(DeviceManagerError::Zbus)
     }
+
+    async fn reboot(&self) -> Result<(), DeviceManagerError> {
+        // DBUS should never stopped before us during reboot, so an error is an actual error here
```
rgallor, line 209: i would rewrite this comment, it's not so clear

rgallor (changes requested):

rgallor (approved):

harlem88 (approved):

PR #510, opened 2025-03-17
chore(release): use "thin" instead of "fat" lto

joshuachp:
This achieves similar performance while taking substantial less time.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,10 +46,11 @@ rust-version = "1.72"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
-[profile.release-minimal]
+[profile.release-optimized]
 inherits = "release"
 strip = true
-lto = true
+# Similar to fat in term of performances, but takes substantially less time to run
+lto = "thin"
 debug = true
 # Supported by all platform
 split-debuginfo = "packed"


harlem88 (approved):

rgallor (approved):
LGTM, we just have to remember to change the profile in other repositories where it is used. Can we document this change somewhere?

joshuachp:
It's not the default profile for release, so it's not impacting users. And it's that way for a good reason.

Anyone using this should know that it wasn't "stable" since we had and possibly still have issues with cross compiling with lto. I don't feel it's stable yet to use it, and so documenting it for use is not advisable.

If someone is opting in to lto, I hope [they](https://doc.rust-lang.org/cargo/reference/profiles.html#lto) [know](https://doc.rust-lang.org/rustc/linker-plugin-lto.html#linker-plugin-based-lto) [the](https://doc.rust-lang.org/rustc/linker-plugin-lto.html#toolchain-compatibility) [drawbacks](https://users.rust-lang.org/t/dramatic-increase-in-compile-time-with-fat-lto-in-release-build-causes-and-troubleshooting/99539) [.](https://github.com/rust-lang/rust/issues?q=is%3Aissue%20state%3Aopen%20%20label%3AA-LTO%20)

harlem88 (approved):

PR #529, opened 2025-05-15
fix(containers): make host.docker.internal work on all systems

joshuachp:


joshuachp:
If needed in the release-0.9 this option can be set in the container runtime.

PR #533, opened 2025-05-28
feat(cross): add feature to make it easier to cross compile 

joshuachp:
Add the `cross` feature and `Cross.toml` to make it easier to cross
compile the device runtime.

### DIFF
--- a/doc/Cross.toml
+++ b/doc/Cross.toml
@@ -1,0 +1,24 @@
+# This file is part of Astarte.
+#
+# Copyright 2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+[target.armv7-unknown-linux-musleabihf]
+pre-build = [
+  # additional commands to run prior to building the package
+  "dpkg --add-architecture armhf",
+  "apt-get update && DEBIAN_FRONTEND=noninteractive apt-get --assume-yes install cmake clang",
+]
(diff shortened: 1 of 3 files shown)


Cross.toml
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+[target.armv7-unknown-linux-musleabi]
+pre-build = [
+  # additional commands to run prior to building the package
+  "dpkg --add-architecture armhf",
+  "apt-get update && DEBIAN_FRONTEND=noninteractive apt-get --assume-yes install crossbuild-essential-armhf cmake clang",
```
harlem88, line 23: Since all other dependencies are vendored, we can add only `clang` to satisfy the bindgen requred by aws-lc-sys
```suggestion
  "apt-get update",
  "apt-get --assume-yes install clang"
```

doc/Cross.toml
```diff
@@ -0,0 +1,6 @@
+[target.armv7-unknown-linux-musleabi]
+pre-build = [
+  # additional commands to run prior to building the package
+  "dpkg --add-architecture armhf",
+  "apt-get update && DEBIAN_FRONTEND=noninteractive apt-get --assume-yes install crossbuild-essential-armhf cmake clang",
```
harlem88, line 5: same above

harlem88 (changes requested):

Cross.toml
joshuachp: I don't think cmake is not installed by default on ubuntu container, so I think it would be better to explicitly install the dependencies to prevent future breakages.

I removed the crossbuild-essential.

doc/Cross.toml
joshuachp: I removed this file

Cross.toml
```diff
@@ -0,0 +1,24 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+[target.armv7-unknown-linux-musleabi]
```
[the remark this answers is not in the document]
joshuachp: No idea, changed

doc/Cross.toml
```diff
@@ -0,0 +1,24 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+[target.armv7-unknown-linux-musleabihf]
```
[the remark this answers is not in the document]
joshuachp: I think we should add them when needed

PR #561, opened 2025-08-13
feat(containers): add extra hosts to CreateContainerRequest 

joshuachp:
This will make it possible to add customs hosts to the /etc/hosts file
inside the container.

In particular, it will be possible to set the
`host.docker.internal:host-gateway` hostname, pointing to  the docker
host.


edgehog-device-runtime-store/src/models/containers/container.rs
```diff
@@ -434,6 +434,22 @@ impl ToSql<Integer, Sqlite> for HostPort {
...
+#[diesel(table_name = crate::schema::containers::container_extra_hosts)]
+#[diesel(belongs_to(Container))]
+#[diesel(check_for_backend(diesel::sqlite::Sqlite))]
+pub struct ContainerExtraHost {
+    /// [`Container`] id
+    pub container_id: SqlUuid,
+    /// Host to add to /etc/hosts inside the container.
+    ///
+    /// Its in the form of Host:IP
+    pub value: String,
```
rgallor, line 450: It seems we don't check if the string is actually in the format `<host>:<IP>`. Do you think that adding this check could be a good idea?

PR #587, opened 2025-09-05
fix(timestamp): add timestamp and improve the error messages 

joshuachp:


harlem88 (approved):

lucaato:
By doing a search for `send_object\(` i still find two occurrences:
- `src/telemetry/wifi_scan.rs:121` i think this still has to be replaced since inside the the `send_wifi_scan` function calls `client.send_object_with_timestamp`
- `src/telemetry/battery_status.rs:267` same thing as above

Have you already checked those ?

joshuachp:
> * `` same thing as above

Those are some expectation in the tests that are unfortunately never run. I will open an issue to mock the Wi-Fi telemetry and battery status methods to actually test them.

lucaato (approved):

PR #620, opened 2025-10-01
fix(containers): improve queries to select deployments for deletion

joshuachp:
Select resources that are not used by other deployments. The previews queries didn't work because rows would not be excluded since they would include the container id.


edgehog-device-runtime-containers/src/resource/container.rs
```diff
@@ -161,6 +161,7 @@ where
     }
 
     async fn delete(&mut self, ctx: &mut Context<'_, D>) -> Result<()> {
+        self.container.stop(ctx.client).await?;
```
harlem88, line 164: Is it right to remove the stop function from here?

joshuachp: I've added the stop function because the backend could send a delete for a running container, and docker would error if we don't stop the container first.

harlem88 (approved):

lucaato (approved):

PR #624, opened 2025-10-03
feat(container/telemetry): integreate with the edgehog runtime telemetry infrastructure

joshuachp:
Use the telemetry interface and configuration file to configure the container telemetry interfaces.


edgehog-device-runtime-containers/src/stats/mod.rs
```diff
@@ -47,148 +52,281 @@ mod volume;
...
-    #[instrument(skip(self))]
-    async fn containers(&mut self) -> eyre::Result<()> {
-        let containers: Vec<ContainerId> = self
+    /// Loads the container ids from the storage
+    async fn load_container_ids(&self) -> Option<Vec<ContainerId>> {
+        let handle = self.get_handle()?;
+
+        let containers: Vec<ContainerId> = handle
             .store
             .load_containers_in_state(vec![ContainerStatus::Stopped, ContainerStatus::Running])
```
lucaato: Unrelated to the changes but this could probably be a simple slice reference.

joshuachp: We need to move them to another thread, so this would either be an array or a Vec, since it needs to be owned.

For now I think a Vec is better, to not use const generics for the arrays

src/telemetry/stats/storage_usage.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// Get structured data for `io.edgehog.devicemanager.StorageUsage` interface.
+    ///
+    /// The `/dev/` prefix is excluded from the device names since it is common for all devices.
+    pub fn parse(disk: &Disk) -> Option<(String, Self)> {
+        let name = disk.name().to_string_lossy();
+        let name = name.strip_prefix("/dev/").unwrap_or(&name);
+
+        // remove disks with a higher depth
+        if name.contains('/') {
+            warn!("not simple disks device, ignoring");
```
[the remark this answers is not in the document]
joshuachp: I will print the name and kind of the disk

edgehog-device-runtime-containers/src/stats/mod.rs
lucaato: `Box::new([...])` should also work, but it doesn't change much, do as you see fit

lucaato (approved):

PR #642, opened 2025-10-28
fix(containers): send events to the service loop

### LINKED ISSUE #643: When asked to delete a running deployment, the Runtime publishes incorrect resource properties after their deletion
## How to reproduce

From the Edgehog's cloud dashboard:
- Deploy an application to the device, wait for it to be ready/acknowledged.
- Start the deployment, wait for it to be `Started` with containers running.
- Delete the deployment.

Here are the messages published by the device:

Image

Both the device and Edgehog cloud successfully clean up resources.

During the deletion of the deployment and its resources, the device:
1. Publishes updated properties on each resource (e.g. Image, Container) to reflect its status while carrying out the deletion process.
2. Unsets properties for resources that completed the deletion process.
3. Publishes new properties about the deployment's resources.

However 3. is not ok since it leaves Astarte interfaces advertising statuses for resources that no longer exist on the device: hence, there is a state drift between Astarte and the device, which can lead to further issues for clients using Astarte to interact with the device.

joshuachp:
This prevents concurrency problems when changing properties during the
deletion of a resource.

Closes #643

### DIFF
--- a/edgehog-device-runtime-containers/src/events/mod.rs
+++ b/edgehog-device-runtime-containers/src/events/mod.rs
@@ -18,25 +18,16 @@
 
 //! Send events to Astarte
 
-use bollard::models::{ContainerStateStatusEnum, EventMessageTypeEnum};
+use bollard::models::EventMessageTypeEnum;
 use eyre::Context;
 use futures::{future, TryStreamExt};
-use tracing::{debug, error, instrument, warn};
+use tokio::sync::mpsc;
+use tracing::{debug, error, instrument};
 use uuid::Uuid;
 
 use crate::{
-    container::ContainerId,
-    image::ImageId,
-    network::NetworkId,
-    properties::{
-        container::{AvailableContainer, ContainerStatus},
-        image::AvailableImage,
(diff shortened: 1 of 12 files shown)


edgehog-device-runtime-containers/src/events/mod.rs
```diff
@@ -108,66 +105,20 @@ where
...
 
         Ok(())
     }
 
     #[instrument(skip(self))]
     async fn handle_image(&mut self, local_id: String) -> eyre::Result<()> {
-        let Some((id, reference)) = self.store.find_image_by_local_id(local_id.clone()).await?
-        else {
+        let Some((id, _)) = self.store.find_image_by_local_id(local_id.clone()).await? else {
             debug!("couldn't find image");
```
harlem88:
```suggestion
            warn!("couldn't find image");
```

joshuachp: In this file we should just emit debug since we are listening to all the container events, not just the one regarding the edgehog device runtime.

harlem88 (approved):

PR #651, opened 2025-11-28
feat(rauc): add option to configure the dbus_socket

joshuachp:
Add the [ota.rauc] option and dbus_socket to chose to use either the system or the session bus to comunicate with RAUC.

### DIFF
--- a/src/ota/rauc.rs
+++ b/src/ota/rauc.rs
@@ -29,8 +29,10 @@ use zbus::proxy;
 use zbus::zvariant::{DeserializeDict, SerializeDict, Type};
 
 use crate::error::DeviceManagerError;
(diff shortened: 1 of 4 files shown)


src/ota/rauc.rs
```diff
@@ -211,8 +213,23 @@ impl SystemUpdate for OTARauc<'static> {
...
+                let conn = zbus::Connection::system().await?;
+
+                info!("connected to the system dbus");
+
+                conn
+            }
+            RaucDbus::User => {
+                let conn = zbus::Connection::session().await?;
+
+                info!("connected to the session dbus");
```
harlem88:
```suggestion
                debug!("connected to the session dbus");
```

PR #660, opened 2025-12-18
chore(v0.9.2): prepare release 

joshuachp:


lucaato (approved):

edgehog-device-runtime-tls/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+
+use rustls::crypto::CryptoProvider;
+use rustls::{ClientConfig, Error, RootCertStore};
+use rustls_native_certs::load_native_certs;
+use tracing::instrument;
+use tracing::{error, info};
+
+/// Returns a configure TLS client.
+#[instrument]
+pub fn config() -> Result<ClientConfig, Error> {
```
lucaato: is there a way to do this once and store it in a `OnceLock` or something similar ? Should we try to do it ?

edgehog-device-runtime-tls/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+            platfrom_verifier(provider)
+        } else if #[cfg(feature = "webpki-roots")]{
+            webpki(provider)
+        } else {
+            _default(provider, roots)
+        }
+    }
+}
+
+#[cfg(feature = "platfrom-verifier")]
```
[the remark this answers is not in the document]
joshuachp: I prefer having the compiler use it's heuristic

edgehog-device-runtime-tls/src/lib.rs
joshuachp: I don't think it's needed, the expensive part is reading the root certs. The caller can create an Arc it wants

PR #688, opened 2026-03-09
feat(fileTransfer): http file download

lucaato:


src/http/mod.rs
```diff
@@ -0,0 +1,381 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum FileTransferError {
+    #[error("requested range can't be satisfied")]
+    RangeNotSatisfiable,
+    #[error("expected range {exp:?} got {got:?}")]
+    UnexpectedRange { exp: Range<u64>, got: Range<u64> },
+    #[error(transparent)]
+    Other(#[from] Report),
+}
```
joshuachp: Do we need the range error variants? Can we return the report?

PR #707, opened 2026-03-24
 feat(fileTransfer): add astarte progress reporting 

lucaato:


src/file_transfer/mod.rs
```diff
@@ -266,20 +293,37 @@ impl<F, S, C> FileTransfer<F, S, C> {
...
+                            if req.progress {
+                                let progress = self.start_track_progress(id, None, rx).await?;
+                                let reader = tokio_util::io::ReaderStream::new(progress);
 
-                            let _ = self.client.upload(&req.url, req.headers, reader).await?;
+                                let _ = self.client.upload(&req.url, req.headers, reader).await?;
+                            } else {
+                                let reader = tokio_util::io::ReaderStream::new(rx);
+                                let _ = self.client.upload(&req.url, req.headers, reader).await?;
+                            }
```
harlem88: What do you think about it?
```suggestion
                            let stream = if req.progress {
                               self.start_track_progress(id, None, rx).await?
                            } else {
                                rx
                            }
                            let reader = tokio_util::io::ReaderStream::new(stream);
                            let _ = self.client.upload(&req.url, req.headers, reader).await?;
```

harlem88: What do you think about it?
```suggestion
                            let stream = if req.progress {
                               self.start_track_progress(id, None, rx).await?
                            } else {
                                rx
                            }
                            let reader = tokio_util::io::ReaderStream::new(stream);
                            let _ = self.client.upload(&req.url, req.headers, reader).await?;
```

harlem88 (changes requested):

src/file_transfer/file_system/store.rs
```diff
@@ -389,6 +389,134 @@ where
...
+        std::task::ready!(this.inner.as_mut().poll_read(cx, buf))?;
+
+        let read = buf.filled().len() - before;
+        let read =
+            u64::try_from(read).map_err(|e| io::Error::new(io::ErrorKind::FileTooLarge, e))?;
+
+        *this.remaining = this.remaining.checked_sub(read).ok_or(io::Error::new(
+            io::ErrorKind::FileTooLarge,
+            "file read limit exceeded",
+        ))?;
```
joshuachp:
```suggestion
        *this.remaining = buf.filled().len().checked_sub(before)
            .and_then(|read| u64::try_from(read).ok())
            .and_then(|read| this.remaining.checked_sub(read))
            .ok_or(io::Error::new(
                io::ErrorKind::FileTooLarge,
                "file read limit exceeded",
            ))?;
```

joshuachp (changes requested):

src/file_transfer/mod.rs
lucaato: I don't think this can be done since the two types are actually different they can't be assigned to the same variable.

lucaato: same as above but i can try to improve the amount of repeated code

src/file_transfer/file_system/store.rs
lucaato: ok I'll merge the expressions in the other trait impls where it's feasible

lucaato:
Simplified a bit the code even if it meant wrapping always the reader/writer in `Progress` let me know what you think.

harlem88 (approved):

joshuachp:
Pls, squash the commits or use conventional commit (the `test():` will not appear in the change log)

src/file_transfer/interface/status.rs
```diff
@@ -120,3 +120,47 @@ fn to_errno(kind: io::ErrorKind) -> i32 {
         _ => 22,                                      // EINVAL (Generic fallback)
     }
 }
+
+#[derive(Debug, Clone, PartialEq, IntoAstarteObject)]
+pub(crate) struct FileTransferProgress {
+    id: String,
+    #[astarte_object(rename = "type")]
+    ty: TransferDirection,
+    // TODO this should be a long integer if we want to send file bytes as progress
```
joshuachp: This can be removed

src/file_transfer/file_system/store.rs
```diff
@@ -389,6 +389,137 @@ where
...
+#[derive(Debug)]
+#[pin_project]
+pub(crate) struct Progress<T, S> {
+    #[pin]
+    inner: T,
+    total: u64,
+    // To avoid blocking we use a watch
+    sender: watch::Sender<S>,
+    // To be generic over the content of the watch provide a function that updates the value
+    update_fn: fn(&mut S, u64),
```
joshuachp: Here you are passing a function pointer, it's probably better to pass a F: FnMut if you want to be generic.

src/file_transfer/interface/status.rs
```diff
@@ -120,3 +120,50 @@ fn to_errno(kind: io::ErrorKind) -> i32 {
...
 }
+
+#[derive(Debug, Clone, PartialEq, IntoAstarteObject)]
+pub(crate) struct FileTransferProgress {
+    id: String,
+    #[astarte_object(rename = "type")]
+    ty: TransferDirection,
+    bytes: i64,
+    total_bytes: i64,
+}
```
joshuachp, line 134: Add a snapshot test on serializing the object

src/file_transfer/interface/status.rs
```diff
@@ -120,3 +120,50 @@ fn to_errno(kind: io::ErrorKind) -> i32 {
...
+    {
+        device
+            .send_object(Self::INTERFACE, "/request", AstarteObject::try_from(self)?)
+            .await
+            .map_err(eyre::Error::from)
+    }
+
+    pub(crate) fn set_bytes(&mut self, bytes: i64) {
+        self.bytes = std::cmp::max(bytes, 0);
+    }
```
joshuachp, line 248: This feels weird, I would pass the u64 and cap it at max i64

joshuachp (changes requested):

src/file_transfer/interface/status.rs
```diff
@@ -120,3 +120,50 @@ fn to_errno(kind: io::ErrorKind) -> i32 {
...
+        }
+    }
+
+    /// Create progress event with total bytes, only positive value are accepted
+    pub(crate) fn with_total(id: FileTransferId, bytes: i64, total_bytes: i64) -> Self {
+        Self {
+            id: id.uuid().to_string(),
+            ty: id.direction(),
+            bytes: std::cmp::max(bytes, 0),
+            total_bytes: std::cmp::max(total_bytes, 0),
```
joshuachp: Same here, pass u64 and cap it to i64 max. Also printing a warning

src/file_transfer/file_system/store.rs
```diff
@@ -389,6 +389,139 @@ where
...
+            .and_then(|read| u64::try_from(read).ok())
+            .and_then(|read| this.remaining.checked_sub(read))
+            .ok_or(io::Error::new(
+                io::ErrorKind::FileTooLarge,
+                "file read limit exceeded",
+            ))?;
+
+        std::task::Poll::Ready(Ok(()))
+    }
+}
```
joshuachp:
```suggestion
```

We don't have limitation in upload

joshuachp (changes requested):

src/file_transfer/interface/status.rs
```diff
@@ -120,3 +120,118 @@ fn to_errno(kind: io::ErrorKind) -> i32 {
...
+    ty: TransferDirection,
+    bytes: i64,
+    total_bytes: i64,
+}
+
+impl FileTransferProgress {
+    const INTERFACE: &str = "io.edgehog.devicemanager.fileTransfer.Progress";
+
+    fn to_i64(unsigned: u64) -> i64 {
+        i64::try_from(unsigned).unwrap_or(i64::MAX)
```
joshuachp:
```suggestion
        i64::try_from(unsigned).inspect_err(|error| warn!(%error, "progress bytes overflow")).unwrap_or(i64::MAX)
```

src/file_transfer/interface/status.rs
```diff
@@ -120,3 +120,118 @@ fn to_errno(kind: io::ErrorKind) -> i32 {
...
+    pub(crate) fn with_total(id: FileTransferId, bytes: u64, total_bytes: u64) -> Self {
+        Self {
+            id: id.uuid().to_string(),
+            ty: id.direction(),
+            bytes: Self::to_i64(bytes),
+            total_bytes: Self::to_i64(total_bytes),
+        }
+    }
+
+    pub(crate) async fn send<C>(self, device: &mut C) -> eyre::Result<()>
```
joshuachp, line 172:
```suggestion
    #[instument(skip_all)]
    pub(crate) async fn send<C>(self, device: &mut C) -> eyre::Result<()>
```

src/file_transfer/mod.rs
```diff
@@ -155,23 +167,32 @@ impl<C> FileTransfer<Fs, SysPipe, C> {
...
         })
     }
 }
 
-impl<F, S, C> FileTransfer<F, S, C> {
+impl<F, S, C> FileTransfer<F, S, C>
+where
+    F: 'static,
+    S: 'static,
+    C: 'static,
```
joshuachp: Could you explain why is this required?

joshuachp (changes requested):

src/file_transfer/mod.rs
lucaato: explained irl, avoided using `use<>` explicit capture as a solution

src/file_transfer/file_system/store.rs
lucaato: used a trait after an irl conversation

src/io/progress.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        this.progress.update(*this.total);
+
+        std::task::Poll::Ready(Ok(()))
+    }
+}
+
+pub(crate) trait ProgressHandle {
+    fn update(&mut self, bytes: u64);
+}
```
joshuachp, line 196:
```suggestion
/// Handle the progress of an IO operation.
pub(crate) trait ProgressHandle {
    /// Callback with the progress in bytes.
    ///
    /// This is the total number of bytes read or written.
    fn update(&mut self, bytes: u64);
}
```

src/io/progress.rs
joshuachp, line 1: Can we add some test for this module?

src/io/progress.rs
```diff
@@ -0,0 +1,124 @@
...
+            ))?;
+
+        this.progress.update(*this.total);
+
+        std::task::Poll::Ready(Ok(()))
+    }
+}
+
+pub(crate) trait ProgressHandle {
+    fn update(&mut self, bytes: u64);
```
joshuachp:
```suggestion
    fn update(&mut self, bytes: u64) -> io::Result<()>;
```

src/file_transfer/mod.rs
```diff
@@ -393,15 +473,68 @@ impl<F, S, C> FileTransfer<F, S, C> {
...
+    }
+}
+
+impl ProgressHandle for watch::Sender<Option<FileTransferProgress>> {
+    fn update(&mut self, bytes: u64) {
+        self.send_modify(|old| {
+            if let Some(old) = old {
+                old.set_bytes(bytes);
+            }
+        });
```
```suggestion
    #[instrument(skip(self))]
    fn update(&mut self, bytes: u64) -> io::Result<()> {
        let mut res = Ok(());
        
        self.send_modify(|old| {
            if let Some(old) = old {
                old.set_bytes(bytes);
            } else {
                warn!("file transfer removed from channel");
                
                res = Err(io::Error::from(io::ErrorKind::BrokenPipe))
            }
        });
        
        res
```

src/file_transfer/mod.rs
```diff
@@ -155,23 +168,27 @@ impl<C> FileTransfer<Fs, SysPipe, C> {
...
         F: Space,
         S: Pipe,
         C: astarte_device_sdk::Client + Send + Sync + 'static,
     {
         self.storage.init().await?;
 
-        // TODO cancel
-        loop {
+        while !cancel.is_cancelled() {
             self.jobs().await?;
```
joshuachp: Cancel this future in the while

joshuachp (changes requested):

src/file_transfer/mod.rs
```diff
@@ -393,15 +454,107 @@ impl<F, S, C> FileTransfer<F, S, C> {
...
+                    if let Some(p) = event {
+                        trace!(?p, "transfer progress event");
+
+                        if let Err(e) = p.send(&mut self.device).await {
+                            error!(err=%e, "error while sending progress to astarte");
+                        }
+                    }
+                }
+            })
+            .await;
```
joshuachp: Here we should cancell only the `rx.changed` call, since the `device.send` is not cancel safe

src/file_transfer/mod.rs
```diff
@@ -393,15 +454,107 @@ impl<F, S, C> FileTransfer<F, S, C> {
...
     }
 }
 
+#[derive(Debug)]
+enum ProgressUpdate {
+    Track {
+        sender: watch::Sender<Option<FileTransferProgress>>,
+    },
+    None,
+}
```
joshuachp: This should be a struct, you're re implementing option this way

joshuachp (changes requested):

joshuachp (approved):

PR #717, opened 2026-03-31
feat(fileTransfer): validate the transfer digest 

joshuachp:


src/file_transfer/digest.rs
```diff
@@ -0,0 +1,266 @@
...
+use std::io::{self};
+use std::task::ready;
+
+use pin_project::pin_project;
+use tokio::io::{AsyncBufReadExt, AsyncRead, AsyncReadExt, AsyncSeek, AsyncWrite, BufReader};
+use tracing::instrument;
+
+use super::request::FileDigest;
+
+/// Wrapper to a [`AsyncRead`] or [`AsyncWrite`] that calculates the written digest.
```
lucaato:
```suggestion
/// Wrapper to a [`AsyncRead`] or [`AsyncWrite`] that calculates the digest.
```

src/io/digest.rs
```diff
@@ -0,0 +1,266 @@
...
+            let mut reader = BufReader::new(inner.take(start));
+
+            while let buf = reader.fill_buf().await?
+                && !buf.is_empty()
+            {
+                ctx.update(buf);
+
+                let len = buf.len();
+
+                reader.consume(len);
```
lucaato, line 65:
```suggestion
                reader.consume(buf.len());
```
nit

src/file_transfer/digest.rs
```diff
@@ -0,0 +1,266 @@
...
+        bufs: &[io::IoSlice<'_>],
+    ) -> std::task::Poll<io::Result<usize>> {
+        let this = self.project();
+
+        for buf in bufs {
+            this.ctx.update(buf);
+        }
+
+        this.inner.poll_write_vectored(cx, bufs)
+    }
```
lucaato: I'm not sure this is correct by looking at the implementation of the [`AsyncWrite::poll_write_vectored`](https://docs.rs/tokio/1.43.0/src/tokio/io/async_write.rs.html#158-168) method. I think we should stick to defining the `poll_write` impl.

src/file_transfer/digest.rs
```diff
@@ -0,0 +1,266 @@
...
+        Self {
+            ctx: digest.into(),
+            inner,
+        }
+    }
+
+    #[instrument(skip(inner))]
+    pub(crate) async fn from_read(inner: S, digest: FileDigest, start: u64) -> io::Result<Self>
+    where
+        S: AsyncRead + AsyncSeek + Unpin,
```
lucaato:
```suggestion
        S: AsyncRead + Unpin,
```
i think that async seek is not needed here

src/file_transfer/file_system/store.rs
```diff
@@ -389,6 +387,39 @@ where
     }
 }
 
+impl<W> AsyncSeek for Limit<W>
```
lucaato: I don't think this is needed currently as it also requires another field in the struct.

lucaato (changes requested):

src/io/digest.rs
joshuachp: This is intentional, the buf and reader are borrowed in the loop

src/file_transfer/digest.rs
joshuachp: sure

src/io/digest.rs
```diff
@@ -0,0 +1,248 @@
...
+use std::io::{self};
+use std::task::ready;
+
+use pin_project::pin_project;
+use tokio::io::{AsyncBufReadExt, AsyncRead, AsyncReadExt, AsyncWrite, BufReader};
+use tracing::instrument;
+
+use super::request::FileDigest;
+
+/// Wrapper to a [`AsyncRead`] or [`AsyncWrite`] that calculates the digest.
```
lucaato, line 29: Should digest be moved to the new `crate::io` module ?

src/file_transfer/mod.rs
```diff
@@ -722,6 +723,11 @@ mod tests {
     #[rstest]
     #[tokio::test]
     async fn should_download_full_when_unsatisfiable() {
+        tracing_subscriber::registry()
+            .with(tracing_subscriber::fmt::layer().with_test_writer())
+            .with(tracing_error::ErrorLayer::default())
+            .init();
```
lucaato:
```suggestion
```

lucaato (changes requested):
If you rebase this and correct the two things i mentioned it can be merged imo

harlem88 (approved):

lucaato (approved):

PR #730, opened 2026-04-09
feat(fileTransfer): add device capabilities

joshuachp:


harlem88 (approved):

src/file_transfer/interface/capabilities.rs
```diff
@@ -0,0 +1,110 @@
...
+
+use crate::data::set_property;
+
+/// File transfer capabilities the Device supports
+pub const INTERFACE: &str = "io.edgehog.devicemanager.fileTransfer.Capabilities";
+/// Capabilities of the device
+pub const CAPABILITIES: Capabilities<1, 2> = Capabilities {
+    encodings: ["tar.gz"],
+    unix_permissions: true,
+    targets: ["storage", "streaming"],
```
lucaato: should we use constants for these values ?

joshuachp: We are already in a constant

joshuachp: Good call 🍖

PR #735, opened 2026-04-14
feat(fileTransfer): add filesystem as a target

joshuachp:
Permit the download and upload to any provided path on the device.


src/file_transfer/mod.rs
```diff
@@ -248,103 +255,148 @@ impl<F, S, C> FileTransfer<F, S, C> {
...
+            .await
+            .wrap_err_with(|| format!("couldn't access file: {}", path.display()))?;
 
-        let pipe = self.stream.open_writer(&opt).await?;
+        if exists {
+            info!("file already exists");
 
-        let total_len = download.download_length();
+            return Ok(());
+        }
```
lucaato, line 359: Should we somehow tell the server that the file was already on the system ?

joshuachp: We will tell that the download is a success, in the next PR I will check if the existing file matches the provided digest.

src/file_transfer/interface/request.rs
```diff
@@ -21,12 +21,23 @@ use tracing::{error, instrument};
...
+use super::status::TransferDirection;
+
 #[derive(Debug, Clone, PartialEq)]
 pub(crate) enum FileTransferRequest {
     Download(ServerToDevice),
     Upload(DeviceToServer),
 }
 
+impl FileTransferRequest {
+    pub(crate) fn resp_id(&self) -> (&str, TransferDirection) {
```
lucaato, line 33:
```suggestion
    pub(crate) fn resp_id(&self) -> FileTransferId {
```
could you use FileTransferId here? It was supposed to have a generic defaulting to Uuid

src/file_transfer/mod.rs
```diff
@@ -232,155 +241,163 @@ impl<F, S, C> FileTransfer<F, S, C> {
...
+        file.finalize(&opt).await?;
+
+        Ok(())
+    }
+
+    async fn download_to_write_handle(
+        &mut self,
+        download: &Download<'_>,
+        file: WriteHandle,
+    ) -> Result<WriteHandle, eyre::Error> {
```
lucaato:
```suggestion
    ) -> eyre::Result<WriteHandle> {
```

src/file_transfer/interface/request.rs
joshuachp: This is used only in one place, I don't think is called for

src/file_transfer/mod.rs
```diff
@@ -232,160 +241,215 @@ impl<F, S, C> FileTransfer<F, S, C> {
...
             .wrap_err("download request error")?;
 
-        // limit to maximum uncompressed file size
-        let total = file_resp.total_length().unwrap_or(download.file_size);
+        let limit_size = file_resp.total_length().unwrap_or(download.file_size);
+        let progress = FileTransferProgress::start(
+            download.transfer(),
+            download.progress,
+            file_resp.total_length(),
+        );
```
lucaato, line 326: to achieve this you could use
```suggestion
        let progress = download.progress.then(FileTransferProgress::start(
            download.transfer(),
            file_resp.total_length(),
        ));
```

src/file_transfer/file_system/mod.rs
```diff
@@ -34,3 +44,193 @@ pub(crate) struct FileOptions {
...
+                    if let Err(error) = std::os::unix::fs::chown(to, None, gid) {
+                        use tracing::error;
+
+                        error!(%error, "couldn't change file ownership, this could be because the user is unprivileged and doesn't belongs to the target group");
+                    }
+                })
+                .await?;
+            } else {
+                let _ = opt;
+            }
```
harlem88, line 144: What do you think about? so you don't need to add `else { let _ = opt; }`
```suggestion
            #[cfg(unix)] 
            {
                let gid = opt.perm.group_id;

                tokio::task::spawn_blocking(move || {
                    if let Err(error) = std::os::unix::fs::chown(to, None, gid) {
                        use tracing::error;

                        error!(%error, "couldn't change file ownership, this could be because the user is unprivileged and doesn't belongs to the target group");
                    }
                })
                .await?;
       }
```

harlem88 (changes requested):

src/file_transfer/file_system/mod.rs
joshuachp: You need to add the `let _opt = opt;` since otherwise the variable is unused on windows.

src/file_transfer/mod.rs
joshuachp: I've tried, but I like the function since reduce the complexity of the code used in two places

PR #737, opened 2026-04-16
refactor(fileTransfer): feature gate the implementation

joshuachp:
For systems that don't require the feature of the file-transfer we want
to be able to disable it, to reduce the resource usage and binary size.


harlem88 (approved):

src/file_transfer/http/mod.rs
```diff
@@ -75,7 +61,7 @@ impl FtHttpClient {
     }
 
     // TODO: Add total if available
```
lucaato:
```suggestion
```
I think this can be removed

lucaato (approved):

PR #738, opened 2026-04-16
feat(fileTransfer): configure the service 

joshuachp:
Add a `file_transfer` section to the configuration file:

- `enabled`: flag to start the file transfer service
- `storage_dir`: directory for the storage destination and source
- `storage_reserved`: percentage of free space to keep for the system

### DIFF
--- a/src/controller/mod.rs
+++ b/src/controller/mod.rs
@@ -48,9 +48,10 @@ pub struct Runtime<T> {
     cancel: CancellationToken,
     telemetry_tx: mpsc::Sender<TelemetryEvent>,
     #[cfg(feature = "file-transfer")]
-    file_transfer: mpsc::Sender<crate::file_transfer::interface::request::FileTransferRequest>,
+    file_transfer:
+        Option<mpsc::Sender<crate::file_transfer::interface::request::FileTransferRequest>>,
     #[cfg(feature = "containers")]
-    containers_tx: mpsc::Sender<Box<edgehog_containers::requests::ContainerRequest>>,
+    containers_tx: Option<mpsc::Sender<Box<edgehog_containers::requests::ContainerRequest>>>,
     #[cfg(feature = "forwarder")]
     forwarder: crate::forwarder::Forwarder<T>,
     #[cfg(all(feature = "zbus", target_os = "linux"))]
@@ -122,9 +123,9 @@ impl<C> Runtime<C> {
         #[cfg(feature = "file-transfer")]
         let file_transfer = Self::file_transfer(
             client.clone(),
+            opts.file_transfer,
             tasks,
             jobs,
-            opts.store_directory.join("file-store"),
             cancel.child_token(),
         )
         .wrap_err("could't initialize file transfer")?;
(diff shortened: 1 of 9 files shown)


src/file_transfer/config.rs
```diff
@@ -0,0 +1,141 @@
...
+pub const DEFAULT_MAX_FREE_PERCENTAGE: Percentage = Percentage::new(20).unwrap();
+
+/// Value from 0 to 100
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Percentage(u8);
+
+impl Percentage {
+    /// Creates a new percentage.
+    pub const fn new(value: u8) -> Option<Self> {
+        if value < 100 { Some(Self(value)) } else { None }
```
lucaato:
```suggestion
        if value <= 100 { Some(Self(value)) } else { None }
```

src/controller/mod.rs
```diff
@@ -192,19 +195,25 @@ impl<C> Runtime<C> {
 
         use self::actor::Persisted;
 
+        if !config.enabled {
+            tracing::debug!("file transfer service not enabled");
```
lucaato: nit this could be at info since it logged just once

src/file_transfer/config.rs
```diff
@@ -0,0 +1,141 @@
...
+
+//! Configuration for the file-transfer service.
+
+use std::ops::Deref;
+use std::path::{Path, PathBuf};
+
+use serde::{Deserialize, Serialize};
+
+/// Max free space to keep on the device.
+pub const DEFAULT_MAX_FREE_PERCENTAGE: Percentage = Percentage::new(20).unwrap();
```
lucaato, line 27: would it be easier, from a user perspective, to declare the max space that we use instead of the max space that we will be leaving free

lucaato: I would also rephrase and write `maximum percentage of available space used by the file transfer`

joshuachp: It think the max free is easier and is also what other programs do. (on top of mind Longhorn kubernetes Storage manager)

The used space could be an additional feature, but we would need to check the used space of the folder. Which i think it's possible (like how the command `du` does it), but it's an additional feature.

If you want you can open an issue for it, I personally feel the free space is sufficient for our needs right now.

joshuachp: Also I think in that case it would be better to set a file transfer max storage size in GiB/MiB not a percentage.

src/controller/mod.rs
```diff
@@ -192,19 +195,25 @@ impl<C> Runtime<C> {
 
         use self::actor::Persisted;
 
+        if !config.enabled {
+            tracing::info!("file transfer service not enabled");
```
harlem88, line 199: What do you think to set as debug level? It was probably the user who disabled it.
```suggestion
            tracing::debug!("file transfer service not enabled");
```

src/controller/mod.rs
```diff
@@ -217,10 +226,16 @@ impl<C> Runtime<C> {
         >,
         tasks: &mut JoinSet<eyre::Result<()>>,
         cancel: CancellationToken,
-    ) -> eyre::Result<mpsc::Sender<Box<edgehog_containers::requests::ContainerRequest>>>
+    ) -> eyre::Result<Option<mpsc::Sender<Box<edgehog_containers::requests::ContainerRequest>>>>
     where
         C: Client + Send + Sync + 'static,
     {
+        if !config.enabled {
+            tracing::info!("container service not enabled");
```
harlem88, line 234:
```suggestion
            tracing::debug!("container service not enabled");
```

src/controller/mod.rs
joshuachp: this was sudgested as https://github.com/edgehog-device-manager/edgehog-device-runtime/pull/738#discussion_r3118726142

joshuachp: I think it's okay since it's printed only once

harlem88: I don't like too many logs in the output, but ok it is just only once .

harlem88 (approved):

lucaato (approved):

PR #739, opened 2026-04-16
 feat(fileTransfer): cleanup downloaded partial file 

lucaato:


src/file_transfer/file_system/mod.rs
```diff
@@ -171,6 +171,20 @@ impl WriteHandle {
...
+    #[instrument(skip_all, fields(self.path))]
+    pub(crate) async fn delete(mut self) -> io::Result<()> {
+        let partial = self.partial;
+
+        info!("flushing and closing file");
+        self.file.flush().await?;
+        drop(self.file);
+
+        info!("removing file");
+        tokio::fs::remove_file(&partial).await?;
```
harlem88: Do we need to take these log at info level?
```suggestion
        debug!("flushing and closing file");
        self.file.flush().await?;
        drop(self.file);

        debug!("removing file");
        tokio::fs::remove_file(&partial).await?;
```

harlem88 (approved):

src/file_transfer/mod.rs
```diff
@@ -1176,6 +1263,54 @@ mod tests {
...
+        //         function(|o: &AstarteObject| {
+        //             let code = o.get("code").unwrap();
+
+        //             code != &AstarteData::Integer(0)
+        //         }),
+        //     )
+        //     .once()
+        //     .in_sequence(&mut seq)
+        //     .returning(|_, _, _| Ok(()));
+        let (mut transfer, dir) = mk_def_transfer("digest_error", device).await;
```
joshuachp, line 1261: Should this be removed?

joshuachp (commented):
I'm not sure this is sufficient, we are not cleaning up the finalize_write in the download_store

src/file_transfer/mod.rs
```diff
@@ -633,6 +677,16 @@ impl ProgressHandle for ProgressUpdate {
     }
 }
 
+trait IntoInner<T> {
+    fn into_inner(self) -> T;
+}
```
joshuachp: This trait is used only here we can implement it on progress

src/file_transfer/mod.rs
lucaato: imo it's ok to leave it i'll use it in the next pr to check that the error is sent to astarte, if you want I can remove it anyways from here

lucaato:
> I'm not sure this is sufficient, we are not cleaning up the finalize_write in the download_store

That should get cleaned up by the finalize_write function, when the file gets decompressed this is done in the function and when it gets moved there shouldn't be any need to do it.

joshuachp:
That's not the case if the archive is invalid or there's an io::Error. AFAICS the file would not be removed.

I think the clean up should be hoisted in the handle function in the download_store and download_file. If needed we can also create a `struct TransferError { error: io::Error, fatal: bool // default true }`.

lucaato:
You're right as per irl conversations I've moved the deletion logic in one place, currently it does not perform any special type of check on the error and deletes the partial file whenever there's an error

src/io/digest.rs
```diff
@@ -72,12 +72,9 @@ impl<S> Digest<S> {
...
+    pub(crate) fn check_digest(self, digest: &[u8]) -> Result<S, S> {
         if self.ctx.finish().as_ref() != digest {
-            return Err(io::Error::new(
-                io::ErrorKind::InvalidInput,
-                "data digest mismatch",
-            ));
+            return Err(self.inner);
         }
 
         Ok(self.inner)
```
joshuachp, line 80:
```suggestion
    pub(crate) fn check_digest(self, digest: &[u8]) -> io::Result<Self> {
        self.ctx.finish().as_ref() != digest {
            return Err(io::Error::new(
                io::ErrorKind::InvalidInput,
                "data digest mismatch",
            ));
        }

        Ok(self.inner)
```

This can be reverted since I don't believe we use S on error anymore

src/file_transfer/file_system/mod.rs
```diff
@@ -94,7 +94,8 @@ impl WriteHandle {
 
         let file = Digest::from_read(file, alg, meta.len()).await?;
 
-        file.into_inner(digest)?;
+        file.check_digest(digest)
+            .map_err(|_| io::Error::new(io::ErrorKind::InvalidInput, "data digest mismatch"))?;
```
joshuachp, line 98:
```suggestion
        file.check_digest(digest)?;
```

For change to check_digest

src/file_transfer/file_system/mod.rs
```diff
@@ -135,29 +136,36 @@ impl WriteHandle {
...
+            self.delete().await?;
+
+            Ok(())
         } else {
             self.move_part(opt).await
         }
     }
 
-    #[instrument(skip_all)]
+    #[instrument(skip_all, fields(self.path))]
```
joshuachp:
```suggestion
    #[instrument(skip_all)]
```

Since it's printed for finalize

src/file_transfer/file_system/mod.rs
```diff
@@ -175,15 +183,15 @@ impl WriteHandle {
         Ok(())
     }
 
-    #[instrument(skip_all)]
-    async fn extract(mut self, compression: Encoding) -> io::Result<()> {
+    #[instrument(skip_all, fields(self.path))]
```
joshuachp:
```suggestion
    #[instrument(skip_all)]
```

Since it's printed for finalize

src/file_transfer/mod.rs
```diff
@@ -316,34 +324,53 @@ impl<F, S, C> FileTransfer<F, S, C> {
...
+                info!("file was already deleated");
+
+                return Ok(());
+            }
+
+            eyre::bail!(error);
+        }
 
         Ok(())
     }
```
```suggestion
        match handle.delete().await {
            Ok(()) => Ok(()),
            Err(err) if err.kind() == io::ErrorKind::NotFound {
                debug!("file already missing");
                
                Ok(())
            }
            Err(error) => {
                error!("couldn't delete the file");
                
                eyre::bail!(error)
            }
        }
}
```

src/file_transfer/mod.rs
```diff
@@ -316,34 +324,53 @@ impl<F, S, C> FileTransfer<F, S, C> {
...
             .into_inner()
             .into_inner()
-            .into_inner(&download.digest)?;
+            .check_digest(&download.digest)
+            .map_err(|_| eyre!("digest mismatch, streamed data may be corrupted"))?;
+
+        Ok(())
+    }
+
+    async fn handle_write_error(handle: WriteHandle) -> eyre::Result<()> {
```
```suggestion
    #[instrument(skip_all)]
    async fn handle_write_error(handle: WriteHandle) -> eyre::Result<()> {
```

joshuachp (changes requested):

src/io/digest.rs
lucaato: Imo it's better overall as a generic api to be able to get the handle back, since it's not really an io::Error but a digest mismatch

joshuachp (approved):

PR #746, opened 2026-04-20
 feat(fileTransfer): schedule file deletion jobs

joshuachp:
Add a unix timestamp schedule_at nullable field to the job queue table
to run a task at a specific instant. Spawn a new task that will wait to
the moment a file need to be cleaned up.


src/file_transfer/mod.rs
```diff
@@ -438,6 +461,30 @@ impl<F, S, C> FileTransfer<F, S, C> {
...
+        let job = Job::try_from(CleanUp {
+            id: download.id,
+            schedule_at: ts.as_i64(),
+            file_path: self.storage.file_path(&download.id).into(),
+        })?;
+
+        if self.queue.insert_if_missing(job).await? {
+            trace!("task notified");
+
+            self.cleanup.notify_one();
```
lucaato, line 482: It's kind of weird that we are always using this queue and the notify together, maybe they should be in the same struct ?

src/jobs/mod.rs
```diff
@@ -125,14 +159,74 @@ impl Queue {
...
+                    .filter(job_queue::schedule_at.is_not_null())
+                    .filter(job_queue::schedule_at.le(now.as_i64()))
+                    .filter(job_queue::job_type.eq(job_type))
+                    .filter(job_queue::status.eq(JobStatus::Pending))
+                    .order_by(job_queue::schedule_at.asc())
+                    .first(write)
+                    .optional()?
+                else {
+                    return Ok(None);
+                };
```
lucaato: nit split into two assignments to improve clarity

src/jobs/mod.rs
```diff
@@ -125,14 +159,74 @@ impl Queue {
...
+        first.map(Unix::try_from).transpose()
+    }
+
+    #[instrument(skip(self))]
+    pub async fn next_scheduled_job(&self, job_type: JobType) -> eyre::Result<Option<Job>> {
+        let now = Unix::now()?;
+
+        let job = self
+            .db
+            .for_read(move |write| {
```
lucaato: Is it correct to use `for_read` even when you are updating?

src/file_transfer/mod.rs
joshuachp: This is already in it's own method and is used in a single place for now

src/jobs/mod.rs
joshuachp: good catch 👍🏻

joshuachp: Added a test for this

src/file_transfer/housekeeping.rs
```diff
@@ -0,0 +1,239 @@
...
+    #[instrument(skip_all)]
+    async fn wait_next(&self, cancel: &CancellationToken) -> eyre::Result<bool> {
+        debug!("waiting for next job");
+
+        if cancel.is_cancelled() {
+            return Ok(false);
+        }
+
+        if let Some(some) = self.queue.next_schedule(JobType::FileStorage).await? {
+            let Some(instant) = some.wait_until()? else {
```
harlem88: nit: use a right name
```suggestion
        if let Some(file_job) = self.queue.next_schedule(JobType::FileStorage).await? {
            let Some(instant) = file_job.wait_until()? else {
```

harlem88 (changes requested):

harlem88 (approved):

lucaato (approved):

PR #748, opened 2026-04-22
chore(fileTransfer): remove partial downloads without job

joshuachp:
Add an extra check on init for partial downloads without a job
associated to them.

### DIFF
--- a/src/file_transfer/file_system/mod.rs
+++ b/src/file_transfer/file_system/mod.rs
@@ -61,7 +61,8 @@ pub(crate) struct WriteHandle {
 }
 
 impl WriteHandle {
-    pub(crate) const PARTTIAL_EXT: &str = ".part";
+    pub(crate) const PARTIAL_EXT: &str = ".part";
+    pub(crate) const PARTIAL: &str = "part";
 
     pub(crate) fn current_size(&self) -> u64 {
         self.current_size
@@ -104,7 +105,7 @@ impl WriteHandle {
     pub(crate) async fn open(path: PathBuf, opt: &FileOptions) -> io::Result<Self> {
         // Set partial extension
         let mut partial = path.clone().into_os_string();
-        partial.push(Self::PARTTIAL_EXT);
+        partial.push(Self::PARTIAL_EXT);
         let partial = PathBuf::from(partial);
 
         let mut file_options = File::options();
--- a/src/file_transfer/file_system/store.rs
+++ b/src/file_transfer/file_system/store.rs
@@ -18,20 +18,25 @@
 
 //! Handles file system operations for the file transfer.
 
+use std::ffi::OsStr;
 use std::fmt::Debug;
 use std::io;
 use std::path::{Path, PathBuf};
 
+use edgehog_store::models::job::job_type::JobType;
 use eyre::WrapErr;
 use futures::{Stream, TryStreamExt};
+use tokio::fs::ReadDir;
 use tokio_stream::wrappers::ReadDirStream;
-use tracing::{instrument, trace};
+use tracing::{error, instrument, trace, warn};
 use uuid::Uuid;
 
 use crate::file_transfer::config::Percentage;
 use crate::file_transfer::encoding::Paths;
 use crate::file_transfer::interface::file::StoredFile;
 use crate::file_transfer::request::FileDigest;
+use crate::file_transfer::request::TransferJobTag;
+use crate::jobs::Queue;
 
 use super::{FileOptions, WriteHandle};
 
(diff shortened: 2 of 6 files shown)


src/file_transfer/file_system/store.rs
```diff
@@ -50,14 +54,75 @@ impl FileStorage<Fs> {
...
     #[instrument(skip_all)]
-    pub(crate) async fn init(&self) -> eyre::Result<()> {
+    pub(crate) async fn init(&self, queue: &Queue) -> eyre::Result<()> {
         tokio::fs::create_dir_all(&self.dir)
             .await
             .wrap_err("couldn't create file storage directory")?;
 
+        let dir = tokio::fs::read_dir(&self.dir)
+            .await
+            .wrap_err("couldn' read file storage directory")?;
```
harlem88: typo : `couldn'` ->`couldn't`

harlem88 (changes requested):

src/file_transfer/file_system/mod.rs
```diff
@@ -62,6 +62,7 @@ pub(crate) struct WriteHandle {
 
 impl WriteHandle {
     const PARTTIAL_EXT: &str = ".part";
+    const PARTTIAL: &str = "part";
```
lucaato:
```suggestion
const PARTIAL_EXT: &str = ".part";
    const PARTIAL: &str = "part";
```

src/file_transfer/file_system/store.rs
```diff
@@ -50,14 +54,75 @@ impl FileStorage<Fs> {
...
+
+            let res = queue
+                .exists(&id, JobType::FileTransfer, TransferJobTag::Download.into())
+                .await;
+
+            let job_exists = match res {
+                Ok(exists) => exists,
+                Err(error) => {
+                    error!(%error, "couldn't check if job exists");
+                    continue;
```
lucaato, line 101: should this fall back to deleting the partial anyway ?

src/file_transfer/file_system/store.rs
```diff
@@ -50,14 +54,75 @@ impl FileStorage<Fs> {
...
+                }
+            };
+
+            // the formatter would join this if with the error condition which is confusing, so i
+            // separated the conditions.
```
lucaato: no need for this comment imo

src/file_transfer/file_system/store.rs
```diff
@@ -50,14 +54,75 @@ impl FileStorage<Fs> {
 
 impl<F> FileStorage<F> {
     #[instrument(skip_all)]
-    pub(crate) async fn init(&self) -> eyre::Result<()> {
+    pub(crate) async fn init(&self, queue: &Queue) -> eyre::Result<()> {
```
lucaato, line 61: logging the dir here could be a good idea if not done elsewhere

src/file_transfer/file_system/store.rs
joshuachp: No, because this could be a problem with our query or the database

src/file_transfer/file_system/store.rs
```diff
@@ -50,14 +54,76 @@ impl FileStorage<Fs> {
...
+            };
+
+            if job_exists {
+                continue;
+            }
+
+            if let Err(error) = tokio::fs::remove_file(&path).await {
+                error!(%error, path = %path.display(), "couldn't remove partial file")
+            }
+        }
```
harlem88, line 112: I think you can optimize this block, What do you think about this?
```suggestion
            let job_exists = queue
            .exists(&id, JobType::FileTransfer, TransferJobTag::Download.into())
            .await
            .map_err(|error| error!(%error, "couldn't check if job exists"))
            .unwrap_or(true); // Default to true on error to avoid deleting files safely

        if !job_exists {
            if let Err(error) = tokio::fs::remove_file(&path).await {
                error!(%error, path = %path.display(), "couldn't remove partial file");
            } else {
                info!(path = %path.display(), "cleaned up orphaned partial file");
            }
        }
```

harlem88 (changes requested):

lucaato (approved):

src/file_transfer/file_system/store.rs
joshuachp: I removed the comment that explained it as per 😢 https://github.com/edgehog-device-manager/edgehog-device-runtime/pull/748#discussion_r3148131558

I started the first impl like yours, but Clippy would warn that the `if !job_exists` and `let Err()` could be merged, and I wanted the `remove_file` as a separate operation.

harlem88 (approved):

PR #749, opened 2026-04-22
feat(fileTransfer): limit the file size to the configured free space

joshuachp:
Check that a file would not surpass the configured reserved free space
on the device.

### DIFF
--- a/src/file_transfer/file_system/store/mod.rs
+++ b/src/file_transfer/file_system/store/mod.rs
@@ -23,6 +23,7 @@ use std::fmt::Debug;
 use std::io;
 use std::path::{Path, PathBuf};
 
+use cfg_if::cfg_if;
 use edgehog_store::models::job::job_type::JobType;
 use eyre::WrapErr;
 use futures::{Stream, TryStreamExt};
@@ -34,11 +35,15 @@ use uuid::Uuid;
 use crate::file_transfer::config::Percentage;
 use crate::file_transfer::encoding::Paths;
 use crate::file_transfer::interface::file::StoredFile;
-use crate::file_transfer::request::FileDigest;
-use crate::file_transfer::request::TransferJobTag;
+use crate::file_transfer::request::{FileDigest, TransferJobTag};
 use crate::jobs::Queue;
 
 use super::{FileOptions, WriteHandle};
+
+#[cfg(unix)]
+mod unix;
+#[cfg(windows)]
+mod windows;
 
 /// Stores files in the storage
 #[derive(Debug)]
@@ -59,7 +64,7 @@ impl FileStorage<Fs> {
 impl<F> FileStorage<F> {
     #[instrument(skip_all)]
     pub(crate) async fn init(&self, queue: &Queue) -> eyre::Result<()> {
-        trace!(path = %self.dir.display(), "initializing store dir");
+        trace!(path = %self.dir.display(), "initialazing store dir");
 
         tokio::fs::create_dir_all(&self.dir)
             .await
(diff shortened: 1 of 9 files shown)


src/file_transfer/file_system/store/mod.rs
```diff
@@ -200,38 +208,106 @@ pub(crate) trait Space {
...
 
         Ok(())
     }
 }
 
+/// Information of a mounted filesystem.
+///
+/// We use bytes to overcome cross system compatibility.
+///
+/// See `man (3) stavfs` for more details.
```
lucaato: Since this is unix spcific move this to the module maybe

harlem88 (approved):

src/file_transfer/file_system/store/windows.rs
```diff
@@ -0,0 +1,80 @@
...
+        .collect();
+
+    let ret = unsafe { GetVolumePathNameW(path.as_ptr(), vol.as_mut_ptr(), BUFF_SIZE) };
+    if ret == 0 {
+        return Err(io::Error::last_os_error());
+    }
+
+    let mut user_avail = 0u64;
+
+    let ret = unsafe { GetDiskFreeSpaceExW(vol.as_ptr(), &mut user_avail, &mut 0, &mut 0) };
```
lucaato: i did not test this but in the docs is written that

> This parameter does not have to specify the root directory on a disk. The function accepts any directory on a disk.

so maybe the first function call to `GetVolumePathNameW` is not needed ?

src/file_transfer/file_system/store/windows.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use crate::file_transfer::file_system::store::FsStat;
+
+pub(crate) fn get_disk_free_space(path: &Path) -> io::Result<FsStat> {
+    // https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+    // Should be 260
+    const BUFF_SIZE: u32 = MAX_PATH + 1;
+
+    let mut vol = [0u16; BUFF_SIZE as usize];
+    let path: Vec<u16> = path
```
lucaato, line 36: should you also limit this path to MAX_PATH ?

src/file_transfer/file_system/store/windows.rs
lucaato:
```suggestion
    let ret = unsafe { GetDiskFreeSpaceExW(vol.as_ptr(), &mut user_avail, std::ptr::null_mut::<u64>(), std::ptr::null_mut::<u64>()) };
```
using `&mut 0` is different than passing a `NULL` i think, we should pass something like `std::ptr::null_mut::<u64>()` since this could disable some calculation done inside the funciton (`ptr::null_mut()` could be enough`)

src/file_transfer/file_system/store/windows.rs
```diff
@@ -0,0 +1,80 @@
...
+    let mut sectors_per_cluster = 0;
+    let mut bytes_per_sector = 0;
+    let mut total_number_of_clusters = 0;
+
+    let ret = unsafe {
+        GetDiskFreeSpaceW(
+            vol.as_ptr(),
+            &mut sectors_per_cluster,
+            &mut bytes_per_sector,
+            &mut 0,
```
lucaato: same as above

src/file_transfer/file_system/store/windows.rs
joshuachp: This passes a pointer to a valid u32 that we imediatelly discard

src/file_transfer/file_system/store/windows.rs
joshuachp: Here we don't need to provide a pointer to an already allocated buffer, so we can use a dynamic allocation.

src/file_transfer/file_system/store/windows.rs
joshuachp: The doc requires the path to be on a disk, not knowing if the path is absolute/relative or without disk prefix like `C:`, we need to check and actually retrieve the disk path

lucaato: passing a null pointer could disable some computation in the function, and i prefer it since we don't need the value

lucaato: maybe its better to directly make the path absolute in rust and then avoid the first call ?

joshuachp: Sure 👍🏻

src/file_transfer/interface/file.rs
```diff
@@ -29,13 +29,25 @@ use eyre::Context;
...
 #[derive(Debug, Clone)]
-pub(crate) struct StoredFile<S = String> {
-    id: S,
+pub(crate) struct StoredFile {
+    id: String,
     path: PathBuf,
     size: i64,
 }
 
-impl StoredFile<String> {
```
lucaato, line 38: This was useful to avoid extra allocations when notifying created files in the `finalize_write`. Are you sure this is worth it ? It wasn't complicating the code all that much.

joshuachp: We always allocate in this function since we need to convert to the AstarteData::String. I moved the allocation before the creation of the struct to remove the generic.

lucaato (approved):

PR #750, opened 2026-04-22
feat(fileTransfer): send to astarte the list of stored files

lucaato:


src/file_transfer/mod.rs
```diff
@@ -633,6 +649,69 @@ impl<F, S, C> FileTransfer<F, S, C> {
 
         Ok(())
     }
+
+    async fn update_stored_files(&mut self, cancel: &CancellationToken) -> eyre::Result<()>
+    where
+        C: Client + PropAccess + Send + Sync + 'static,
+    {
+        let mut cleanup = self.fetch_previous_files().await?;
+        let mut files = pin!(self.storage.files().await?);
```
joshuachp, line 657: You don't need to pin here, add unpin to the impl Stream in the files return type

src/file_transfer/mod.rs
```diff
@@ -633,6 +649,69 @@ impl<F, S, C> FileTransfer<F, S, C> {
...
+
+        // TODO cleanup interface if we get a parsing error or can't strip
+        let set: Result<HashSet<Uuid>, _> = previous
+            .iter()
+            .filter_map(|p| {
+                p.path
+                    // TODO replace with constant
+                    .strip_suffix("/pathOnDevice")
+                    .and_then(|request_id| request_id.strip_prefix("/"))
+                    .map(Uuid::parse_str)
```
joshuachp: Use string here so we can keep track of random files added to the store

src/file_transfer/file_system/store.rs
```diff
@@ -116,15 +119,36 @@ impl<F> FileStorage<F> {
...
+
+    #[instrument(skip_all)]
+    pub(crate) async fn files(&self) -> io::Result<impl Stream<Item = io::Result<StoredFile>>> {
+        let read_dir = tokio::fs::read_dir(&self.dir).await?;
+
+        let stream = ReadDirStream::new(read_dir).and_then(async |e| {
+            let id = Uuid::parse_str(&e.file_name().to_string_lossy())
+                .map_err(|e| io::Error::new(io::ErrorKind::InvalidFilename, e))?;
+
+            let path = e.path();
```
joshuachp: Also here we should probably use str to keep track of random files

src/file_transfer/interface/file.rs
```diff
@@ -0,0 +1,166 @@
...
+use std::path::PathBuf;
+
+use tracing::instrument;
+use uuid::Uuid;
+
+const INTERFACE: &str = "io.edgehog.devicemanager.storage.File";
+
+#[derive(Debug, Clone)]
+pub(crate) struct StoredFile {
+    id: Uuid,
```
joshuachp: Same here, use a String

src/file_transfer/interface/file.rs
```diff
@@ -0,0 +1,166 @@
...
+        device
+            .unset_property(Self::INTERFACE, &format!("/{}/pathOnDevice", self.id))
+            .await?;
+
+        device
+            .unset_property(Self::INTERFACE, &format!("/{}/sizeBytes", self.id))
+            .await
+            .map_err(eyre::Error::from)
+    }
+}
```
joshuachp: I think this could be a method `File::unset(id: impl<Display>)` on the File property

src/file_transfer/interface/file.rs
```diff
@@ -0,0 +1,166 @@
...
+            size: super::to_i64(size),
+        }
+    }
+
+    #[instrument(skip_all)]
+    pub(crate) async fn send<C>(self, device: &mut C) -> eyre::Result<()>
+    where
+        C: astarte_device_sdk::Client + Send + Sync + 'static,
+    {
+        let path = self.path.display().to_string();
```
joshuachp:
```suggestion
        let path = String::from(self.path.to_str_lossy());
```

To make sure the conversion is always the same

joshuachp (changes requested):
I think we can be flexible on the files in the store directory and use strings instead of uuids

src/file_transfer/mod.rs
lucaato: doesn't work because there is a closure there

src/file_transfer/file_system/store.rs
```diff
@@ -116,15 +119,41 @@ impl<F> FileStorage<F> {
...
+
+            if !name.ends_with(WriteHandle::PARTTIAL_EXT) {
+                let path = e.path();
+                let size = e.metadata().await?.len();
+
+                Ok(Some(StoredFile::create(name, path, size)))
+            } else {
+                Ok(None)
+            }
+        });
```
joshuachp, line 155:
```suggestion
        let stream = ReadDirStream::new(read_dir).try_filter_map(async |e| {
            let name = e.file_name().to_string_lossy();

            if !name.ends_with(WriteHandle::PARTTIAL_EXT) {
                let path = e.path();
                let size = e.metadata().await?.len();

                Ok(Some(StoredFile::create(name.into(), path, size)))
            } else {
                Ok(None)
            }
```

The into converts Cow to String and is a nop for Cow::Owned, while to_string allocates since is from &self

src/file_transfer/interface/file.rs
```diff
@@ -0,0 +1,204 @@
...
+            )
+            .await
+            .map_err(eyre::Error::from)
+    }
+
+    pub(crate) fn id(&self) -> &S {
+        &self.id
+    }
+
+    pub(crate) async fn deleted<C>(id: S, device: &mut C) -> eyre::Result<()>
```
joshuachp: This function should not error, we'll update the properties next reboot

joshuachp (commented):
Looks good, one minor thing and a second where I thinks better to not error while dealing with properties

joshuachp (approved):

PR #751, opened 2026-04-23
chore(fileTransfer): fix query to update status to the correct job 

joshuachp:
Update the single job status to inprogress with the correct query when
getting the next job.

Before we where using diesel Identifiable incorrectly and ending up
updating multiple jobs at once.

### DIFF
--- a/src/jobs/mod.rs
+++ b/src/jobs/mod.rs
@@ -27,7 +27,7 @@ use edgehog_store::models::job::job_type::JobType;
 use edgehog_store::models::job::status::JobStatus;
 use edgehog_store::schema::runtime::job_queue;
 use eyre::Context;
-use tracing::{debug, instrument};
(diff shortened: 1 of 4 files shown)


src/jobs/mod.rs
```diff
@@ -158,19 +158,26 @@ impl Queue {
...
                     .first(write)
-                    .optional()?
-                else {
+                    .optional()?;
+
+                let Some(mut job) = optional else {
                     return Ok(None);
                 };
 
                 job.status = JobStatus::InProgress;
```
lucaato, line 172: Is this neededd since you chanced the below query ?

joshuachp: yes, it doesn't return the new job

lucaato (approved):

harlem88 (approved):

PR #766, opened 2026-05-06
fix(fileTransfer): require content length for known size uploads

lucaato:


src/file_transfer/stream.rs
```diff
@@ -0,0 +1,458 @@
...
+
+    #[tokio::test]
+    async fn pipe_writer() {
+        let mut buf = Vec::new();
+        let write_content = b"ciaociao";
+
+        write_pipe(write_content, &mut buf).await;
+
+        with_insta!({
+            insta::assert_debug_snapshot!(buf);
```
joshuachp: Use the Hexdump here

src/file_transfer/stream.rs
```diff
@@ -0,0 +1,458 @@
...
+impl PipeDataHeader {
+    fn new(length: u64) -> Self {
+        Self { length }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, minicbor::Encode, minicbor::Decode, minicbor::CborLen)]
+pub struct PipeDataFooter {
+    #[cbor(n(0))]
+    status: i32,
```
joshuachp, line 89: Shouldn't this be a u32?

joshuachp (changes requested):

scripts/file-transfer/create-upload-url
joshuachp, line 1: Pls restore this. You can add rustup to the devShell

lucaato: my bad i was supposed to revert the change

PR #777, opened 2026-05-14
feat(fileTransfer): add gz capability where applicable

lucaato:


joshuachp (changes requested):
LGTM. Please sign the commit so we can merge

PR #781, opened 2026-05-19
feat(fileTransfer): add stored file deletion

lucaato:


src/storage/interface.rs
```diff
@@ -0,0 +1,152 @@
...
+    aggregation = "object",
+    path = "/request",
+    rename_all = "camelCase"
+)]
+#[astarte_object(rename_all = "camelCase")]
+pub struct DeleteFile {
+    pub(crate) id: String,
+    pub(crate) file_id: String,
+    pub(crate) force: bool,
+}
```
joshuachp, line 42: Add a test to de serialize the event/interface/astarteObject.

src/storage/interface.rs
```diff
@@ -0,0 +1,152 @@
...
+
+#[derive(Debug, Clone, PartialEq, IntoAstarteObject)]
+pub(crate) struct FileStorageResponse {
+    id: String,
+    #[astarte_object(rename = "type")]
+    ty: ActionType,
+    code: i32,
+    messages: Vec<String>,
+}
```
joshuachp, line 93: Same here, a test for the serialization

src/controller/mod.rs
```diff
@@ -121,14 +123,30 @@ impl<C> Runtime<C> {
 
         // TODO: Add configuration
```
joshuachp: Sorry, not directly inherent but we now have configuration for the file transfer

joshuachp (changes requested):
Looks great some test to add to prevent future breakages and it's good to merge

src/storage/mod.rs
```diff
@@ -0,0 +1,389 @@
...
+            .await?;
+
+        Ok(())
+    }
+
+    async fn delete(&mut self, job: request::Delete) -> eyre::Result<()>
+    where
+        C: Client + Send + Sync + 'static,
+    {
+        let path = self.storage.file_path(&job.id);
```
joshuachp:
```suggestion
        let path = self.storage.file_path(&job.file_id);
```

joshuachp (changes requested):
Please add a test of a file deletion

joshuachp (approved):

PR #782, opened 2026-05-20
refactor(containers): keep track of the order of the deployment's containers

joshuachp:
Add an index column to the `deployment_containers` table to order the contaienr_ids as the one received in the CreateDeployment interface from Edgehog.


joshuachp:
This also introduces a restart for the previous deployment after a failed update

edgehog-device-runtime-containers/src/service/mod.rs
```diff
@@ -529,7 +529,7 @@ impl<D> Service<D> {
     where
         D: Client + Send + Sync + 'static,
     {
-        for id in deployment.containers {
+        for id in deployment.containers.values().copied() {
```
lucaato: Should the deletion follow a reversed order with respect to how they were created ?

joshuachp: Good idea

joshuachp: I don't think it actually changes anything since the containers are already stopped. But I've added the `.rev()` since seems cleaner

lucaato (approved):

harlem88 (approved):
Add just a line to the changelog

harlem88 (approved):

PR #783, opened 2026-05-21
feat(securityEvents): add tracing of security events for containers

lucaato:


edgehog-device-runtime-containers/src/events/deployment.rs
```diff
@@ -115,6 +118,21 @@ impl From<EventStatus> for AstarteData {
...
 
+#[cfg(feature = "security-events")]
+impl crate::tracing::IntoSecurityEvent for &DeploymentEvent {
+    fn into_event(self) -> Option<crate::tracing::SecurityEvent> {
+        let event = match self.status {
+            EventStatus::Starting => crate::tracing::SecurityEvent::ContainerDeploymentInit,
+            EventStatus::Started => crate::tracing::SecurityEvent::ContainerDeploymentOk,
+            EventStatus::Stopping => crate::tracing::SecurityEvent::ContainerUndeploymentInit,
+            EventStatus::Stopped => crate::tracing::SecurityEvent::ContainerUndeploymentOk,
+            _ => return None,
```
joshuachp: this is missing the other events like Error -> DeploymentFail

Cargo.toml
```diff
@@ -95,6 +95,8 @@ bindgen = ["dep:aws-lc-rs", "aws-lc-rs/bindgen", "edgehog-store/bindgen"]
 # Feature to make it easier to cross compile the runtime, it requires clang and cmake to be
 # installed
 cross = ["vendored", "bindgen"]
+# Enable logging of some events to journald
+security-events = ["dep:tracing-journald", "edgehog-containers/security-events", "edgehog-service/security-events"]
```
joshuachp:
```suggestion
security-events = ["dep:tracing-journald", "edgehog-containers?/security-events", "edgehog-service?/security-events"]
```

This will not activate the optional dependency

src/main.rs
```diff
@@ -167,6 +160,35 @@ async fn main() -> stable_eyre::Result<()> {
...
 
+#[cfg(feature = "security-events")]
+fn init_tracing() -> eyre::Result<()> {
+    use tracing_subscriber::Layer;
+
+    let layer = tracing_subscriber::fmt::layer()
+        .with_ansi(std::io::stdout().is_terminal())
+        .with_filter(
+            EnvFilter::builder()
+                .with_default_directive("edgehog_device_runtime=info".parse()?)
```
joshuachp: Can we change this to `tracing::Level::INFO` (on all spans)? Since it's in most of the case what we want to know

src/main.rs
```diff
@@ -167,6 +160,35 @@ async fn main() -> stable_eyre::Result<()> {
...
+
+    let subscribers = tracing_subscriber::registry().with(layer);
+
+    #[cfg(feature = "security-events")]
+    let subscribers = subscribers.with(
+        tracing_journald::layer()?
+            .with_syslog_identifier("edgehog_runtime_security_events".to_string())
+            .with_filter(
+                tracing_subscriber::filter::Targets::new()
+                    .with_target("edgehog-security-event", tracing::Level::TRACE),
```
joshuachp: Can we add the SDK also?

edgehog-device-runtime-containers/src/resource/container.rs
```diff
@@ -41,31 +41,52 @@ impl ContainerResource {
...
+            return e.into();
+        }
 
-        Err(ResourceError::Missing {
+        ResourceError::Missing {
             id: ctx.id,
             resource: "container",
-        })
+        }
     }
```
joshuachp: This seems a complication from the original code, since you removed all the `?` with the `if let` returns.

edgehog-device-runtime-containers/src/events/deployment.rs
```diff
@@ -49,6 +49,9 @@ impl DeploymentEvent {
     where
         D: Client + Sync + 'static,
     {
+        #[cfg(feature = "security-events")]
+        crate::tracing::notify(&self);
```
joshuachp: I would prefer the explicit security event on the action and not in this function. Since it can send duplicate events when we refresh the status of a container. But let's discuss about it.

joshuachp (changes requested):
Looks good. I would always logs these events instead of only with the feature active.

edgehog-device-runtime-containers/src/resource/container.rs
lucaato: It makes clear that the function is returning an error instead of a result, i'll change it a bit

src/main.rs
```diff
@@ -167,6 +160,37 @@ async fn main() -> stable_eyre::Result<()> {
     Ok(())
 }
 
+#[cfg(feature = "security-events")]
```
joshuachp:
```suggestion
```

Cargo.toml
```diff
@@ -95,6 +95,8 @@ bindgen = ["dep:aws-lc-rs", "aws-lc-rs/bindgen", "edgehog-store/bindgen"]
 # Feature to make it easier to cross compile the runtime, it requires clang and cmake to be
 # installed
 cross = ["vendored", "bindgen"]
+# Enable logging of some events to journald
```
joshuachp:
```suggestion
# Enable logging of security events to journald
```

src/main.rs
```diff
@@ -167,6 +162,33 @@ async fn main() -> stable_eyre::Result<()> {
     Ok(())
 }
 
+fn init_tracing() -> eyre::Result<()> {
+    let layer = tracing_subscriber::fmt::layer()
+        .with_ansi(std::io::stdout().is_terminal())
```
joshuachp:
```suggestion
        .with_ansi(!cfg!(windows) && std::io::stdout().is_terminal())
```

joshuachp (changes requested):

joshuachp (approved):

PR #788, opened 2026-05-26
fix(securityEvents): report all errors during deployment operations

lucaato:


joshuachp (changes requested):
We should add the ContainerStop in the stop_deployment function and ContainerStart in the start_deployment function

joshuachp (approved):

PR #789, opened 2026-05-26
fix(securityEvents): move events to match cloud behaviour

lucaato:


edgehog-device-runtime-containers/src/service/mod.rs
joshuachp:
```suggestion

AvailableDeployment::new(&deployment_id)
.send(
&mut self.device,
crate::properties::deployment::DeploymentStatus::Started,
)
.await
.map_err(ResourceError::Property)?;

        crate::tracing::notify(crate::tracing::SecurityEvent::ContainerStartOk);
```

joshuachp: Otherwise will print Container Start OK and StartFail

joshuachp:
```suggestion

AvailableDeployment::new(&deployment)
.send(
&mut self.device,
crate::properties::deployment::DeploymentStatus::Stopped,
)
.await
.map_err(ResourceError::from)?;

        crate::tracing::notify(crate::tracing::SecurityEvent::ContainerStopOk);
```

Same as before

joshuachp (commented):
The update is missing the DeploymnetInit and Ok/Fail

harlem88 (approved):

PR #792, opened 2026-05-29
fix(securityEvents): attach container names as event fields

lucaato:
The container events start/stop in all variants (init/fail/ok) now include `F_CONTAINER_NAME` field that indentifies the container.
When the event is related to a deployment the field `F_DEPLOYMENT_ID` will identify the edgehog deployment.

### DIFF
--- a/edgehog-device-runtime-containers/src/docker/container.rs
+++ b/edgehog-device-runtime-containers/src/docker/container.rs
@@ -48,6 +48,7 @@ use crate::{
         container::{parse_port_binding, RestartPolicy},
         BindingError,
     },
+    tracing::{container_event, ContainerSecurityEvent},
 };
 
 /// Error for the container operations.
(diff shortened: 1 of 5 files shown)


edgehog-device-runtime-containers/src/docker/container.rs
```diff
@@ -221,21 +222,32 @@ impl ContainerId {
     pub(crate) async fn start(&self, client: &Client) -> Result<Option<()>, ContainerError> {
         debug!("starting {self}");
 
+        let container_name = self.container();
```
joshuachp: This is the docker container_id (from the container runtime) or container_name (the id from edgehog). Is it okay to log both?

edgehog-device-runtime-containers/src/service/mod.rs
```diff
@@ -592,12 +601,14 @@ impl<D> Service<D> {
     }
 
     /// Will update an application between deployments
-    #[instrument(skip(self))]
+    #[instrument(target = "edgehog-security-event", skip_all, fields(from_deployment_id=%bundle.from, deployment_id=%bundle.to))]
```
joshuachp: Do we need the target here? Since it's already in the log functions?

edgehog-device-runtime-containers/src/docker/container.rs
lucaato: I think we should only log the docker `container_id` and the edgehog `deployment_id` since it's the value that can group containers id together. But I'm not sure this is correct since this is not specified in the document we received.

edgehog-device-runtime-containers/src/service/mod.rs
lucaato: Basically yes if we want the span fields to be logged to journald together with the message fields. I think this is the cleanest way of doing it since it makes tracing handle it automatically, it's true it changes the logging too.

edgehog-device-runtime-containers/src/docker/container.rs
joshuachp: Here I would print the container ID passed by edgehog, since the other ID is unknown in the frontend

joshuachp (changes requested):

edgehog-device-runtime-containers/src/tracing.rs
```diff
@@ -73,31 +95,74 @@ impl SecurityEvent {
...
-            | Self::ContainerStartFail
-            | Self::ContainerStopFail => Level::ERROR,
+            Self::ContainerStartFail | Self::ContainerStopFail => Level::ERROR,
             _ => Level::INFO,
         }
     }
 }
 
+/// This function returns a span guard that should be used to instrument container events.
+pub fn deployment_span(deployment_id: Uuid) -> tracing::Span {
```
joshuachp, line 105: I believe this needs to be a macro to get the correct span information on where it's called. It's a minor thing since we use it only for the security events you can decide if you want to change it.

joshuachp (approved):

PR #815, opened 2026-06-30
refactor(storageUsage): update the interface

joshuachp:
Change the parameter to use the disk UUID and add the additional fields of the mount paths, kind and name of the disk.


src/telemetry/stats/storage_usage/mod.rs
```diff
@@ -84,10 +81,173 @@ impl TelemetryTask for StorageUsage {
...
+            total_bytes: 50_000_000_000,
+            free_bytes: 25_000_000_000,
+        };
+
+        let exp = AstarteObject::from_iter(
+            [
+                ("name", AstarteData::String("/dev/sda1".to_string())),
+                ("kind", AstarteData::String("HDD".to_string())),
+                ("fstype", AstarteData::String("ext4".to_string())),
+                ("mounts", AstarteData::StringArray(vec!["/".to_string()])),
```
lucaato, line 175: why you didn't `.clone` here ?

joshuachp: I just copy pasted the fields to be quick writing the test

harlem88:
Maybe for some versions, Do we need to support both version of the storageUsage interface?

joshuachp:
Those are additional fields, so it's not a breaking change. The data is still correct/useful only viewing the old fields.

harlem88 (approved):

src/telemetry/stats/storage_usage.rs
```diff
@@ -84,10 +81,182 @@ impl TelemetryTask for StorageUsage {
...
+    cfg_if::cfg_if! {
+        if #[cfg(all(feature = "udev", target_os = "linux"))] {
+            let dev = tokio::fs::metadata(disk.name())
+                .await
+                .inspect_err(|error| error!(%error, "couldn't get device number"))
+                .ok()?;
+
+            return Some(read_udev_uuid_prop(&dev));
+        } else if #[cfg(windows)] {
+            return Some(read_udev_uuid_prop(disk));
```
lucaato: even if branch specific give these functions different names. Just for the sake of clarity

joshuachp: renamed and separated into modules

lucaato (approved):

PR #819, opened 2026-07-06
feat(forwarder): move protos in the workspace

joshuachp:
This will make it possible to update the dependencies and rust-version of the protos in sync with the edgehog-device-runtime.


joshuachp:
This PR start using submodules, so any checkout needs to init and update them

rgallor (approved):

PR #839, opened 2026-07-24
feat(container): add aditional fields to the container interface

joshuachp:


edgehog-device-runtime-store/src/db.rs
```diff
@@ -288,6 +288,9 @@ impl Manager {
 
             if reader {
                 conn.batch_execute("PRAGMA query_only = ON;")?;
+            } else {
+                // See: https://sqlite.org/pragma.html#pragma_optimize
+                conn.batch_execute("PRAGMA optimize=0x10002;")?;
```
lucaato, line 293: the linked page suggests also running `pragma optimize;` periodically. Should we do that too ?

joshuachp: yes, in addition to this we should run it once a day.

I will do it in another PR (also in the SDK), since now it's possible because we spawn additional tokio tasks, and can have one for hosekeeping

lucaato (approved):

PR #860, opened 2026-08-05
refactor(interfaces): bundle the interfaces in the binary 

joshuachp:
This will not require to include the interfaces repo in the edgehog
installation, and the correct interfaces are added based on the features
the binary is compiled with.

### DIFF
--- a/src/data/interfaces.rs
+++ b/src/data/interfaces.rs
@@ -1,0 +1,294 @@
+// This file is part of Edgehog.
+//
+// Copyright 2026 SECO Mind Srl
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


src/data/interfaces.rs
```diff
@@ -0,0 +1,253 @@
...
+    include_str!("../../deps/interfaces/io.edgehog.devicemanager.StorageUsage.json");
+const SYSTEM_INFO: &str =
+    include_str!("../../deps/interfaces/io.edgehog.devicemanager.SystemInfo.json");
+const SYSTEM_STATUS: &str =
+    include_str!("../../deps/interfaces/io.edgehog.devicemanager.SystemStatus.json");
+const CONFIG_TELEMETRY: &str =
+    include_str!("../../deps/interfaces/io.edgehog.devicemanager.config.Telemetry.json");
+
+/// Add the enabled interfaces
+pub(crate) fn add_interfaces<C, S>(
```
lucaato, line 43: these blocks could be refactored into functions

joshuachp: I've done it, but I'm not sure it's easier to use now. It also duplicates all the `cfg!`

lucaato (approved):

PR #897, opened 2026-09-14
fix(fileTransfer): do not try to cleanup file system files

lucaato:


joshuachp (approved):
LGTM

I would change the title of the PR and the commit for the changelog in something like: do not try to cleanup file system files.

PR #901, opened 2026-09-15
feat(fileTransfer): store downloaded file with received name

lucaato:


src/storage/mod.rs
```diff
@@ -268,10 +268,14 @@ impl<C> StorageTask<C> {
...
+
             tokio::fs::remove_dir_all(&file_path)
                 .await
                 .wrap_err("couldn't remove directory")?;
         } else {
+            trace!(path = %file_path.display(), "deleting file");
+
             tokio::fs::remove_file(&file_path)
                 .await
                 .wrap_err("couldn't remove file")?;
```
joshuachp: Both branches can be simplified with a remove_dir_all. Since now we also need to remove the directory.

src/file_transfer/file_system/store/mod.rs
```diff
@@ -68,59 +69,210 @@ impl FileStorage<Fs> {
 }
 
 impl<F> FileStorage<F> {
```
joshuachp, line 72: Please move the migration in it's own module

joshuachp (changes requested):

src/file_transfer/interface/file.rs
```diff
@@ -48,6 +47,36 @@ impl StoredFile {
...
+
+    fn size_endpoint(id: &str) -> String {
+        format!("/{}{}", id, Self::SIZE_ENDPOINT)
+    }
+
+    fn endpoint_id(path: &str) -> Option<&str> {
+        trace!(%path, "extracting id");
+
+        path.split('/').nth(1)
+    }
```
joshuachp, line 62: this is not always correct, I would use this method only in the sync

src/file_transfer/interface/file.rs
```diff
@@ -48,6 +47,36 @@ impl StoredFile {
...
+        C: PropAccess,
+    {
+        let stored = device
+            .property(Self::INTERFACE, &Self::path_endpoint(id))
+            .await?
+            .ok_or_eyre("no path property found")?;
+
+        let AstarteData::String(path) = stored else {
+            bail!("unexpected data type");
+        };
```
joshuachp: Use the String:try_from() method on the AstarteData

src/file_transfer/interface/file.rs
joshuachp: You need to have the parent directory, convert the string to Path, and strip the prefix from the property path.

src/file_transfer/interface/file.rs
```diff
@@ -102,35 +126,42 @@ impl StoredFile {
...
 
-    fn file_name_from_path<'a>(path: &'a str) -> Option<Cow<'a, str>> {
-        let path = Path::new(path);
-        let file_name = path.file_name();
+    pub(crate) async fn update_path<C, S>(id: S, path: &Path, device: &mut C)
+    where
+        S: std::fmt::Display,
+        C: astarte_device_sdk::Client + Send + Sync + 'static,
+    {
+        let path = path.to_string_lossy().to_string();
```
joshuachp: The previous code for file_name_from_path handled the non utf-8 string case

joshuachp: Here will not error and send a path with utf-8 character that cannot be converted back into a path

src/file_transfer/file_system/store/mod.rs
```diff
@@ -68,59 +69,210 @@ impl FileStorage<Fs> {
...
-                .await;
+            let dest = self.dir.join(uuid.to_string());
+
+            if tokio::fs::try_exists(&dest).await.is_ok_and(|b| b) {
+                // NOTE if the destination entry already exists it's possible the migration was
+                // interrupted before moving the file inside the temporary directory.
+                // We skip the move since we could overwrite the file with an empty directory.
+                debug!(dest = %dest.display(), "destination entry already exists skipping move");
+                continue;
+            }
```
joshuachp: I believe that we still want to move the file in this case, to complete the migration

lucaato: no this line specifically is finishing an old migration left unfinished, we could overwrite a file with an empty directory if we move the directory in this case.
if the file is still in it's place the directory will still be created and moved afterwards

src/file_transfer/interface/file.rs
lucaato: this should change only if we change the path of the property which should not happen this often

joshuachp: This accounts for `/store/c0b8c4a9-9e5c-4729-84f8-c99a48d5ed4d`, but not `/store/c0b8c4a9-9e5c-4729-84f8-c99a48d5ed4d/bar`

src/file_transfer/interface/file.rs
lucaato: ok i'll error if this happens

src/file_transfer/interface/file.rs
lucaato: there was an error, I should have used the interface path not the value of the property, this way it's stable and always the first component

src/file_transfer/file_system/store/mod.rs
```diff
@@ -69,58 +69,93 @@ impl FileStorage<Fs> {
...
+        if job_exists {
+            return true;
         }
+
+        if let Err(error) = tokio::fs::remove_file(path).await {
+            error!(%error, path = %path.display(), "couldn't remove partial file")
+        }
+
+        return true;
     }
```
joshuachp: can we merge those functions into a should_cleanup_file?

src/file_transfer/file_system/store/mod.rs
```diff
@@ -69,58 +69,93 @@ impl FileStorage<Fs> {
...
 
-            if let Err(error) = tokio::fs::remove_file(&path).await {
-                error!(%error, path = %path.display(), "couldn't remove partial file")
+                return true;
             }
+        };
+
+        if job_exists {
+            return true;
         }
```
joshuachp:
```suggestion
        if job_exists {
            return false;
}
```

src/file_transfer/file_system/store/mod.rs
joshuachp: Since the continues and stuff make the two functions complicated and we can have a single remove.

src/file_transfer/file_system/store/mod.rs
```diff
@@ -139,18 +174,35 @@ impl<F> FileStorage<F> {
...
+        let partial_file_name = format!("{}{}", id, WriteHandle::PARTIAL_EXT);
+        self.dir.join(partial_file_name)
+    }
+
+    pub(crate) fn file_path(&self, id: &Uuid, name: &Path) -> PathBuf {
+        let name_str = name.as_os_str().to_str();
+
+        if name_str.is_none_or(|s| s.is_empty()) {
+            // when the name is empty use the default migration file name
+            self.dir_path(id).join(WriteHandle::MIGRATION_FILE_NAME)
```
joshuachp: We can rename the MIGRATION_FILE_NAME into a default name

joshuachp (changes requested):

src/file_transfer/request/download.rs
```diff
@@ -228,7 +228,10 @@ impl TryFrom<Job> for Download<'_> {
 #[repr(u8)]
 pub(crate) enum Destination<'a> {
     #[n(0)]
-    Storage,
+    Storage {
+        #[n(0)]
+        name: Cow<'a, Path>,
```
joshuachp: Should this be an option?

src/file_transfer/request/download.rs
```diff
@@ -342,7 +362,7 @@ mod tests {
     }
 
     #[rstest]
-    #[case(("storage", ""), Destination::Storage)]
+    #[case(("storage", "testfile.txt"), Destination::Storage { name: Path::new("testfile.txt").into() })]
```
joshuachp: This should be an option

src/file_transfer/file_system/mod.rs
```diff
@@ -81,6 +82,8 @@ impl WriteHandle {
         alg: FileDigest,
         digest: &[u8],
     ) -> io::Result<bool> {
+        trace!(path = %path.display(), "testing if file exists");
```
joshuachp: this is already printed, you can configure the instrumentation if it's too noisy, like:

```rust
#[instrument(skip_all, fields(path = %path.display()))]

File: src/file_transfer/file_system/mod.rs [hunk 13]
```diff
@@ -119,22 +122,29 @@ impl WriteHandle {
...
+        Self::open(file_path, partial, opt).await
+    }
+
+    #[instrument(skip(opt))]
+    pub(crate) async fn open(
+        path: PathBuf,
+        partial: PathBuf,
+        opt: &FileOptions,
+    ) -> io::Result<Self> {
+        trace!(path = %partial.display(), "opening partial for writing and reading");
```
joshuachp: same as above

src/file_transfer/interface/file.rs
```diff
@@ -48,6 +44,34 @@ impl StoredFile {
...
+
+    fn size_endpoint(id: &str) -> String {
+        format!("/{}{}", id, Self::SIZE_ENDPOINT)
+    }
+
+    fn endpoint_id(path: &str) -> Option<&str> {
+        trace!(%path, "extracting id");
+
+        path.split('/').nth(1)
+    }
```
joshuachp:
```suggestion
    #[instrument]
    fn endpoint_id(path: &str) -> Option<&str> {
        trace!("extracting id");

        path.split('/').nth(1)
    }
```

src/file_transfer/interface/file.rs
```diff
@@ -48,6 +44,34 @@ impl StoredFile {
...
+
+    fn size_endpoint(id: &str) -> String {
+        format!("/{}{}", id, Self::SIZE_ENDPOINT)
+    }
+
+    fn endpoint_id(path: &str) -> Option<&str> {
+        trace!(%path, "extracting id");
+
+        path.split('/').nth(1)
```
joshuachp: split_once('/')
```suggestion
        path.split_once('/').map(|(a, _)| a)
```

joshuachp (changes requested):

src/file_transfer/file_system/store/mod.rs
```diff
@@ -159,33 +211,78 @@ impl<F> FileStorage<F> {
...
+
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidData,
+                "expected a single file in store dir, found > 1",
+            ));
+        }
+
+        Ok(inner.path())
+    }
+
```
joshuachp: Move the impl to the tests module

src/file_transfer/file_system/store/mod.rs
```diff
@@ -219,17 +316,30 @@ impl<F> FileStorage<F> {
...
 
-            if !name.ends_with(WriteHandle::PARTIAL_EXT) {
-                let path = e.path();
-                let size = e.metadata().await?.len();
-
-                Ok(Some(StoredFile::create(name.into(), path, size)))
-            } else {
-                Ok(None)
+            // ignore partial files
+            if name.ends_with(WriteHandle::PARTIAL_EXT) {
```
joshuachp: can we use an os string here?

joshuachp (changes requested):

src/file_transfer/file_system/store/mod.rs
```diff
@@ -159,33 +211,78 @@ impl<F> FileStorage<F> {
...
+
+        Ok(inner.path())
+    }
+
     #[instrument(skip(self))]
-    pub(crate) async fn open_read(&self, id: &Uuid) -> eyre::Result<tokio::fs::File> {
-        let path = self.file_path(id);
+    pub(crate) async fn open_read(&self, id: &Uuid, name: PathBuf) -> io::Result<tokio::fs::File> {
+        let file_path = tokio::fs::canonicalize(&name).await?;
+        let dir_path = self.dir_path(id);
```
joshuachp: we can assume the path is absolute for now and not canonicalize it

joshuachp (changes requested):

src/file_transfer/request/mod.rs
```diff
@@ -224,7 +224,8 @@ mod tests {
         pub(crate) fn target(&self) -> String {
             match self {
                 Request::Download(download) => match &download.destination {
-                    download::Destination::Storage => String::new(),
+                    // target for storage is ignored in these tests
+                    download::Destination::Storage { .. } => String::new(),
```
joshuachp: I think we can return the destination here

src/file_transfer/mod.rs
```diff
@@ -302,7 +302,7 @@ impl<F, S, C> FileTransfer<F, S, C> {
         S: Pipe,
     {
         match &req.destination {
-            Destination::Storage => self.download_store(req).await,
+            Destination::Storage { name } => self.download_store(name.clone().into(), req).await,
```
```suggestion
            Destination::Storage { name } => self.download_store(PathBuf::from(&name), req).await,
```

joshuachp (changes requested):

src/file_transfer/file_system/store/mod.rs
lucaato: I kept them separated but the delete is now in a single place, it seems easy to understand now

src/file_transfer/interface/file.rs
lucaato: these are not equal, we have 2 '/' and we need only the middle segment

src/file_transfer/file_system/store/mod.rs
lucaato: i need to do it if i want to check that the path received is inside the uuid target directory (i think i should), i should probably also do it for the store dir

src/file_transfer/interface/file.rs
joshuachp: Your right 👍🏻

src/file_transfer/file_system/store/mod.rs
lucaato: i need to do it to check that a property path is not outside it's uuid directory

src/file_transfer/file_system/store/mod.rs
lucaato: i don't think that would change the behavior that much and i don't see a method to do it easily

src/file_transfer/request/download.rs
lucaato: I think it should always have a value, and when it doesn't as it will be in old edgehog instances the empty path will be handled as well. I guess it comes down to personal preferences since it does not change much. Should i change it ?

